PR #3, opened 2025-02-11
factor_m_dense

thowell:
`mjx.factor_m` with dense Cholesky factorization and tile operations.

### DIFF
--- a/mujoco/mjx/__init__.py
+++ b/mujoco/mjx/__init__.py
@@ -6,6 +6,6 @@ from ._src.io import put_data
 from ._src.smooth import kinematics
 from ._src.smooth import com_pos
 from ._src.smooth import crb
-from ._src.smooth import factor_m, factor_m_dense
+from ._src.smooth import factor_m
 from ._src.test_util import benchmark
 from ._src.types import *
--- a/mujoco/mjx/__init__.py
+++ b/mujoco/mjx/__init__.py
@@ -6,6 +6,6 @@ from ._src.io import put_data
 from ._src.smooth import kinematics
 from ._src.smooth import com_pos
 from ._src.smooth import crb
-from ._src.smooth import factor_m
+from ._src.smooth import factor_m, factor_m_dense
 from ._src.test_util import benchmark
 from ._src.types import *
--- a/mujoco/mjx/__init__.py
+++ b/mujoco/mjx/__init__.py
@@ -7,5 +7,6 @@ from ._src.smooth import kinematics
 from ._src.smooth import com_pos
 from ._src.smooth import crb
 from ._src.smooth import factor_m
+from ._src.support import is_sparse
 from ._src.test_util import benchmark
 from ._src.types import *
--- a/mujoco/mjx/_src/io.py
+++ b/mujoco/mjx/_src/io.py
@@ -1,9 +1,14 @@
 import warp as wp
 import mujoco
 import numpy as np
-import jax
 
 from . import types
+
+
+def _is_sparse(m: mujoco.MjModel):
+  if m.opt.jacobian == mujoco.mjtJacobian.mjJAC_AUTO:
+    return m.nv >= 60
+  return m.opt.jacobian == mujoco.mjtJacobian.mjJAC_SPARSE
 
 
 def put_model(mjm: mujoco.MjModel) -> types.Model:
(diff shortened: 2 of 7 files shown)


mujoco/mjx/__init__.py
```diff
@@ -6,6 +6,6 @@
 from ._src.smooth import kinematics
 from ._src.smooth import com_pos
 from ._src.smooth import crb
-from ._src.smooth import factor_m
+from ._src.smooth import factor_m, factor_m_dense
```
erikfrey, line 9: I like to group imports like this for legibility of files that people will read (even though pystyle and other stuff will complain if you do), but no one really reads __init__ files so much, so it's OK here to just do one import per line.

mujoco/mjx/_src/io.py
```diff
@@ -115,6 +118,23 @@ def put_data(mjm: mujoco.MjModel, mjd: mujoco.MjData, nworld: int = 1) -> types.
...
+  adr = 0
+  for i in range(mjm.nv):
+    j = i
+    while j >= 0:
+      qM_dense[i, j] = mjd.qM[adr]
+      qM_dense[j, i] = mjd.qM[adr]
+      j = mjm.dof_parentid[j]
+      adr += 1
+
+  qLD_dense = np.array(jax.scipy.linalg.cho_factor(qM_dense)[0])
```
erikfrey, line 136: can we use numpy linalg cholesky instead?

In fact, I kind of want to remove jax as a dep from here... maybe.  Let me think about that.  But let's try to avoid JAX for now.

mujoco/mjx/_src/io.py
```diff
@@ -133,8 +153,10 @@ def put_data(mjm: mujoco.MjModel, mjd: mujoco.MjData, nworld: int = 1) -> types.
   d.cinert = wp.array(tile_fn(mjd.cinert), dtype=types.vec10, ndim=2)
   d.cdof = wp.array(tile_fn(mjd.cdof), dtype=wp.spatial_vector, ndim=2)
   d.crb = wp.array(tile_fn(mjd.crb), dtype=types.vec10, ndim=2)
-  d.qM = wp.array(tile_fn(mjd.qM), dtype=wp.float32, ndim=2)
-  d.qLD = wp.array(tile_fn(mjd.qLD), dtype=wp.float32, ndim=2)
+  d.qM = wp.array(tile_fn(qM), dtype=wp.float32, ndim=2)
```
erikfrey, line 156: lets reuporse the same qM and qLD for either dense or sparse

for dense, the shape should be (nv, nv), for sparse it should be (1, nM)

mujoco/mjx/_src/smooth.py
```diff
@@ -258,3 +258,19 @@ def qLDiag_div(m: types.Model, d: types.Data):
     wp.launch(qLD_acc, dim=(d.nworld, size), inputs=[m, d, adr])
   
   wp.launch(qLDiag_div, dim=(d.nworld, m.nv), inputs=[m, d])
+
+
+def factor_m_dense(m: types.Model, d: types.Data, block_dim: int = 256):
```
erikfrey, line 263: we don't need to do it in this PR, but we'll eventually want a model opt that's kinda like `model.opt.jacobian` but for mass matrix format ( tassa@ is already on board with adding this as an option to mujoco xml spec )

No action for now.

erikfrey (changes requested):
Looks great!  Just a few nits.

mujoco/mjx/_src/smooth.py
erikfrey, line 263: let's default block_dim = 32 for now until we have a better idea?

mujoco/mjx/__init__.py
thowell: `factor_m_dense` is removed from the imports because `factor_m` now calls `_factor_m_sparse` or `_factor_m_dense` based on the `Model` field `is_sparse` .

mujoco/mjx/_src/io.py
thowell: done

mujoco/mjx/_src/io.py
```diff
@@ -115,6 +118,23 @@ def put_data(mjm: mujoco.MjModel, mjd: mujoco.MjData, nworld: int = 1) -> types.
   # TODO(erikfrey): would it be better to tile on the gpu?
   tile_fn = lambda x: np.tile(x, (nworld,) + (1,) * len(x.shape))
 
+  # sparse
+  qM = mjd.qM
+  qLD = mjd.qLD
+
+  # dense
+  qM_dense = np.zeros((mjm.nv, mjm.nv))
```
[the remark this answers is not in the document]
thowell: done

mujoco/mjx/_src/smooth.py
thowell: done

erikfrey (approved):
Looks great.

PR #7, opened 2025-02-13
accumulate in local register before saving to memory

erikfrey:
Raises rne perf on my 4090 from 74M -> 98M.


thowell (approved):
nice! lgtm

PR #8, opened 2025-02-14
passive

erikfrey:
Spring and damper passive forces.

### DIFF
--- a/mujoco/mjx/_src/passive_test.py
+++ b/mujoco/mjx/_src/passive_test.py
@@ -1,0 +1,52 @@
+"""Tests for passive force functions."""
+
+from absl.testing import absltest
+from absl.testing import parameterized
+from etils import epath
+import mujoco
+from mujoco import mjx
+import numpy as np
+import warp as wp
+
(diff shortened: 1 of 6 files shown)


mujoco/mjx/_src/passive_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _assert_eq(a, b, name):
+  tol = _TOLERANCE * 10  # avoid test noise
+  err_msg = f'mismatch: {name}'
+  np.testing.assert_allclose(a, b, err_msg=err_msg, atol=tol, rtol=tol)
+
+
+class PassiveTest(parameterized.TestCase):
+
+  def _load(self, fname: str, is_sparse: bool = True):
```
thowell, line 23: should model loading for tests be a utility function?

thowell (approved):
lgtm

mujoco/mjx/_src/passive_test.py
erikfrey: yes agreed, let me do that in a separate PR today as I'm going to add a few more test fixture xmls for params that don't exist in humanoid

PR #18, opened 2025-02-18
Run formatter and fix leftover occurrence of tile_fn in io.py

adenzler-nvidia:
I got a lot of formatting changes on my euler/advance work when running ruff so I think it's better to do a clean sweep of format + isort and then build on top of that.

Also fixes an issue in io.py that made the tests fail on main.

I had to add/change the copyright headers according to the guidance we got from legal, hope that's fine.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,44 @@
+# Copyright 2025 The Physics-Next Project Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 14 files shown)


erikfrey:
Thanks @adenzler-nvidia - the io `tile_fn` issue is fixed in #16

We'll set up CI soon so we can ensure main is always in good shape.

Would love to add ruff linting as well, and a single PR to clean up all the formatting SGTM

adenzler-nvidia:
updated with github action workflow and new license headers. I didn't really test the github actions though, so this needs a good review :)

adenzler-nvidia:
I verified the github action by attempting to merge to main on my fork - there it runs and fails on formatting errors. I'm not sure why it does not run here, maybe a permission issue. We could also just submit and then see whether it starts running.

erikfrey (approved):
Looks good!  I'm going to go ahead and merge since you're hopefully asleep.

PR #20, opened 2025-02-19
initial stab at collision fields

erikfrey:
DO NOT MERGE - WIP

Here's my thinking:

User will ultimately have three parameters that control the size of `Data`:

1. nworld: number of distinct worlds in `Data`
2. nconmax: max number of contacts shared across all worlds in `Data`
3. njmax: max number of constraints shred across all worlds in `Data`

These are optional parameters to `make_data` and `put_data` and we should be able to come up with some reasonable defaults.  We may choose fairly greedy defaults (JAX preallocates like 75% of GPU memory anyway right?)

In order to track number of contacts, we introduce three arrays:
1. `contact` which is the contact metadata, and includes a pointer back to both the world id and the first constraint row id
2. `ncon` which is like MuJoCo, but is batched, so it contains number of contacts per each world
3. `total_ncon` which is a size 1 array, that tracks the total number of contacts across all worlds.  `total_ncon` can be > `nconmax` after a physics step.  This indicates an overflow occurred and some contacts were lost.

### DIFF
--- a/mujoco/mjx/_src/types.py
+++ b/mujoco/mjx/_src/types.py
@@ -22,6 +22,7 @@ class Model:
   nsite: int
   nmocap: int
   nM: int
+  nconmax: int
   opt: Option
   qpos0: wp.array(dtype=wp.float32, ndim=1)
   qpos_spring: wp.array(dtype=wp.float32, ndim=1)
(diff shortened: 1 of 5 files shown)


mujoco/mjx/_src/types.py
```diff
@@ -64,9 +65,32 @@ class Model:
...
+   solref: wp.vec2
+   solreffriction: wp.vec2
+   solimp: wp.types.vector(length=5, dtype=wp.float32)
+   # TODO(team): should dim be int8?
+   dim: wp.int32
+   # TODO(team): should exclude be int8?
+   exclude: wp.int32
+   geom: wp.vec2i
+   efc_address: wp.int32
+   worldid: wp.int32
```
adenzler-nvidia, line 84: this looks ok for the moment, but we might want to look into linearizing the structs into individual arrays. That will help GPU access patterns a lot and should be a lot more performant if we're not processing these in random order (which we shouldn't)

mujoco/mjx/_src/types.py
```diff
@@ -64,9 +65,32 @@ class Model:
...
+
+
 @wp.struct
 class Data:
   nworld: int
+  nconmax: int
+  ncon: wp.array(dtype=wp.int32, ndim=1)
+  # TODO(team): is there a way to express a device scalar or just leave this array(length=1)?
+  ncon_total: wp.array(dtype=wp.int32, ndim=1)  # warp only
```
adenzler-nvidia, line 93: I'm not aware of a better way to do this than single-element arrays.

mujoco/mjx/_src/types.py
```diff
@@ -64,9 +65,32 @@ class Model:
...
+   efc_address: wp.int32
+   worldid: wp.int32
+
+
 @wp.struct
 class Data:
   nworld: int
+  nconmax: int
+  ncon: wp.array(dtype=wp.int32, ndim=1)
+  # TODO(team): is there a way to express a device scalar or just leave this array(length=1)?
```
nvtw, line 92: I think using length 1 arrays to bring scalars to the gpu and back is fine. At least I'm not aware of other specialized solutions for this.

nvtw:
@erikfrey  Would it be possible to merge that guiding structure for the collision_driver from this MR soon because it is required to integrate the collision/contact detection functions?

erikfrey:
Ah yes, this was done in another PR.  Closing this one now.

PR #25, opened 2025-02-21
fwd_actuation, (transmission)

eric-heiden:
Adding `fwd_transmission` for stateless actuators.
I assume `m.actuator_gainprm` and `m.actuator_biasprm` to only have a single value per actuator, which is different from MJX.

I just saw @thowell also implemented a `transmission` function (which is needed for `fwd_actuation`), so we can use his.
I noticed one difference to my implementation for the `transmission` function:
* I'm using `dtype=wp.spatial_vector` for `m.actuator_gear` because it is a 6D vector and offers `wp.spatial_top(v)`, `wp.spatial_bottom(v)` accessors for first and second `vec3` halves.


mujoco/mjx/_src/forward_test.py
```diff
@@ -20,6 +20,8 @@
 import numpy as np
 import warp as wp
 
+wp.config.verify_cuda = True
```
eric-heiden, line 23: This will spit out the name of the kernel as soon as CUDA errors are encountered, maybe we can leave this in for the tests?

.gitignore
```diff
@@ -1,8 +1,9 @@
 # Exclude temporary folders
 *.egg-info/
 env/
+.history
 
 # Python byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
-*$py.class
+*$py.class
```
thowell, line 9: nit: new line at the end

mujoco/mjx/__init__.py
```diff
@@ -17,6 +17,7 @@
 
 from ._src.forward import euler
 from ._src.forward import forward
+from ._src.forward import fwd_actuation
```
thowell, line 20: import transmission

mujoco/mjx/_src/forward_test.py
```diff
@@ -59,6 +61,19 @@ def test_fwd_velocity(self):
     )
     _assert_eq(d.qfrc_bias.numpy()[0], mjd.qfrc_bias, "qfrc_bias")
 
+  def test_fwd_actuation(self):
+    """Tests MJX fwd_actuation."""
+    _, mjd, m, d = self._load("humanoid/humanoid.xml", is_sparse=False)
```
thowell, line 66: set `ctrl` to nonzero values.

mujoco/mjx/_src/math.py
```diff
@@ -43,6 +44,18 @@ def axis_angle_to_quat(axis: wp.vec3, angle: wp.float32) -> wp.quat:
   return wp.quat(c, axis[0], axis[1], axis[2])
 
 
+@wp.func
+def quat_to_axis_angle(quat: wp.quat):
```
thowell, line 48: this function is not used?

mujoco/mjx/_src/smooth.py
```diff
@@ -433,6 +433,76 @@ def qfrc_bias(m: Model, d: Data, cfrc: wp.array(dtype=wp.spatial_vector, ndim=2)
...
+          moment[worldid, actid, vadr + 2] = gear[2]
+          moment[worldid, actid, vadr + 3] = gearaxis[0]
+          moment[worldid, actid, vadr + 4] = gearaxis[1]
+          moment[worldid, actid, vadr + 5] = gearaxis[2]
+        else:
+          for i in range(6):
+            moment[worldid, actid, vadr + i] = gear[i]
+      elif jnt_typ == wp.static(JointType.BALL.value):
+        q = wp.quat(qpos[qadr + 0], qpos[qadr + 1], qpos[qadr + 2], qpos[qadr + 3])
+        # axis, angle = math.quat_to_axis_angle(q)
```
thowell, line 478: remove

mujoco/mjx/_src/math.py
```diff
@@ -14,6 +14,7 @@
 # ==============================================================================
 
 import warp as wp
+from typing import Tuple
```
thowell, line 17: Tuple is not used?

thowell:
for `actuator_gainprm` and `actuator_biasprm` we should follow [MuJoCo convention](https://mujoco.readthedocs.io/en/latest/APIreference/APIglobals.html#glnumeric) where each actuator has `mjNGAIN`=10 and `mjNBIAS`=10 elements.

mujoco/mjx/_src/math.py
eric-heiden: Removed it

mujoco/mjx/_src/forward_test.py
eric-heiden: That was a silly mistake. I've now set some nonzero `ctrl` and `act` values in `ForwardTest._load()` to be used in all forward tests.

eric-heiden:
Thanks for your comments @thowell! I think I've addressed all of them now, including making `actuator_gainprm` and `actuator_biasprm` two-dimensional.

mujoco/mjx/_src/forward_test.py
erikfrey: yeah that's a cool idea

mujoco/mjx/_src/forward.py
```diff
@@ -241,6 +241,64 @@ def _actuator_velocity(d: Data):
...
+      r = m.actuator_ctrlrange[dofid]
+      c = wp.clamp(c, r[0], r[1])
+    f = gain * c + bias
+    if m.actuator_forcelimited[dofid]:
+      r = m.actuator_forcerange[dofid]
+      f = wp.clamp(f, r[0], r[1])
+    force[worldid, dofid] = f
+
+  wp.launch(
+    _force, dim=[d.nworld, m.nu], inputs=[m, d.ctrl], outputs=[d.actuator_force]
```
erikfrey, line 273: OOC so I can write better warp code:

- does sending in `ctrl` rather than all of `Data` help with perf somehow?
- does specifying `outputs=[...` help?  So far I just passed everything in as input!

mujoco/mjx/_src/types.py
```diff
@@ -54,6 +56,43 @@ class Option:
   disableflags: int
 
 
+class TrnType(enum.IntEnum):
```
erikfrey, line 59: Can you chat with @adenzler-nvidia and decide what enum format you guys would like to use?

I'd slightly prefer `IntEnum` but don't care strongly enough to impose, just want you all to be on the same page.

Not blocking for this PR

erikfrey:
OK merging because this generally looks good, please let us know what you want to do about enum types async.

mujoco/mjx/_src/forward.py
eric-heiden: I don't think it makes a noticeable difference to pass in only `ctrl` rather than `Data`. There is some Python overhead for processing arguments when passing multiple arrays rather than one struct of arrays, but that should go away completely with graph captures.

I just specify the outputs separately so we can use `wp.Tape.visualize()` to get some graphviz renders of the computation graph which could be useful for debugging gradients later on: https://nvidia.github.io/warp/modules/differentiability.html#visualizing-computation-graphs

mujoco/mjx/_src/types.py
adenzler-nvidia: I'm happy to follow Eric's lead here - IntEnum looks cleaner and more future proof. I'll prepare a PR for that constants I defined and can check the rest of the code as well.

PR #27, opened 2025-02-21
add xfrc_accumulate

jprozorova:
Adds a xfrc_accumulate function
Adds xfrc_applied field to the Data
Tests that results match with mujoco


mujoco/mjx/_src/support.py
```diff
@@ -1,7 +1,134 @@
...
+    jacr_torque = (
+        jacr_vec[0] * xfrc_applied_vec[3] +
+        jacr_vec[1] * xfrc_applied_vec[4] +
+        jacr_vec[2] * xfrc_applied_vec[5]
+    )
+    
+    wp.atomic_add(qfrc_total[worldid], dofid, jacp_force + jacr_torque)
+
+
+def xfrc_accumulate(m: Model, d: Data):
```
thowell, line 88: it should be possible to process all Cartesian force/torque elements in parallel.

thowell:
please update [fwd_acceleration](https://github.com/erikfrey/mjx_warp/blob/547a2f76d29022388ce95728dcff60790bc9a6ed/mujoco/mjx/_src/forward.py#L248).

mujoco/mjx/_src/support.py
```diff
@@ -1,7 +1,134 @@
...
+
+    # multiply forces with precomputed jcap and jcar to get qfrc
+    qfrc_total = wp.zeros((d.nworld, m.nv), dtype=float)
+    wp.launch(
+        kernel=compute_qfrc,
+        dim=(d.nworld, m.nbody, m.nv),
+        inputs=[d, jacp, jacr, qfrc_total]
+    )
+
+    return qfrc_total
```
thowell, line 134: nit: add a new line here

mujoco/mjx/_src/support.py
```diff
@@ -1,7 +1,134 @@
...
+        kernel=compute_jacobians,
+        dim=(d.nworld,  m.nbody, m.nv),
+        inputs=[d,
+            m,
+            mask, 
+            jacp, 
+            jacr]
+    )
+
+    # multiply forces with precomputed jcap and jcar to get qfrc
```
thowell, line 126: jcap -> jacp, jcar -> jacr

mujoco/mjx/_src/support.py
thowell, line 88: add return type

mujoco/mjx/_src/support.py
```diff
@@ -1,7 +1,134 @@
...
+
+@wp.kernel
+def compute_qfrc(
+    d:Data,
+    jacp:  wp.array3d(dtype=wp.vec3),   
+    jacr: wp.array3d(dtype=wp.vec3),  
+    qfrc_total: wp.array2d(dtype=float)
+):
+    worldid, bodyid, dofid = wp.tid()
+    jacp_vec = jacp[worldid, bodyid, dofid]
```
erikfrey, line 69: MuJoCo splits up `jacp` and `jacr` because it imagined you might want to calculate one or the other, but since we always calculate both (which is fine, simpler that way!) then I think we can use a `spatial_vector` instead of separate `jacp` and `jacr`

mujoco/mjx/_src/support.py
```diff
@@ -1,7 +1,134 @@
...
+        jacp_vec[2] * xfrc_applied_vec[2] 
+    )
+    
+    jacr_torque = (
+        jacr_vec[0] * xfrc_applied_vec[3] +
+        jacr_vec[1] * xfrc_applied_vec[4] +
+        jacr_vec[2] * xfrc_applied_vec[5]
+    )
+    
+    wp.atomic_add(qfrc_total[worldid], dofid, jacp_force + jacr_torque)
```
erikfrey, line 85: We should not need `atomic_add` here since we are threading on dofid, right?

erikfrey (changes requested):

mujoco/mjx/_src/support.py
erikfrey: oh wait nevermind, i see why the `atomic_add` is necessary as you're summing up all bodies into a single dofid

erikfrey (commented):
This is good to start, but besides my comments:

I'm not 100% convinced we need to launch threads over the cartesian product of (nworld x nbody x ndof) - that's potentially a lot of threads!

But I have been surprised by GPU performance properties in the past!

As you make the requested changes, can you please run:

```mjx-testspeed --function=xfrc_accumulate --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192```

And let us know what you're seeing?  Thanks

mujoco/mjx/_src/support.py
erikfrey: wait what's the return type??

thowell:
@erikfrey do you think it is possible to precompute the mask for each body in `put_model` (it looks like this computation is only dependent on `Model`) and then compute all Cartesian force/torque elements in parallel when `xfrc_accumulate` is called?

erikfrey:
@thowell yes possibly could precompute, but let's see the perf impact.  The mask is a teeny tiny matrix.  Also at some point down the line we'll want to do fancy model batching where each Model possibly has different dof trees, so all things even better to not precompute.

BUT if it makes a significant difference to performance, then yeah can slap it on Model.

jprozorova:
@erikfrey  in the original implementation (with requested changes as well) on Ada 600, the total simulation time was 0.07s. By modifying the approach to launch compute_qfrc over (nworld x ndof) threads and looping through the bodies within the kernel I could eliminate the need for atomic additions. The total sim time is brought down to 0.03s.  
With precomputed mask, the time for computing qfrc can be further reduced to 0.02s.

These are all the numbers for current version:
 Total JIT time: 0.26 s
 Total simulation time: 0.03 s
 Total steps per second: 253,631,203
 Total realtime factor: 1,268,156.02 x
 Total time per step: 0.00 µs

erikfrey (approved):
Fantastic, thank you.  This looks great.  I'll leave it to resolve the conflicts and then merge it yourself!  Or if you can't merge just ping me and I'll do it.

PR #28, opened 2025-02-21
solve

thowell:
initial benchmarking results:
(note: the setup is missing the collision pipeline and we should confirm the testspeed setup is correct--we shouldn't read too much into these results for now)

```
Warp 1.7.0.dev20250224 initialized:
   CUDA Toolkit 12.8, Driver 12.2
   Devices:
     "cpu"      : "CPU"
     "cuda:0"   : "NVIDIA RTX A4000" (16 GiB, sm_86, mempool enabled)
```

CG solver
```
mjx-testspeed --function=forward --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192 --nefc_total=32768 --solver=cg --iterations=5 --ls_iterations=5
```
```

 Total JIT time: 57.65 s
 Total simulation time: 8.45 s
 Total steps per second: 969,288
 Total realtime factor: 4,846.44 x
 Total time per step: 1.03 µs
```

Newton solver
```
mjx-testspeed --function=forward --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192 --nefc_total=32768 --solver=newton --iterations=1 --ls_iterations=5
```
```

 Total JIT time: 106.89 s
 Total simulation time: 5.34 s
 Total steps per second: 1,532,808
 Total realtime factor: 7,664.04 x
 Total time per step: 0.65 µs
 ```

### CODE REVIEW

**Comment by erikfrey** [comment]
@thowell Can you add some benchmarking stats?

File: mujoco/mjx/_src/smooth.py [hunk 1]
```diff
@@ -320,7 +320,7 @@ def _factor_i_dense(m: Model, d: Data, M: wp.array, L: wp.array):
   """Dense Cholesky factorizaton of inertia-like matrix M, assumed spd."""
 
   # TODO(team): develop heuristic for block dim, or make configurable
-  block_dim = 32
+  block_dim = 256
```
adenzler-nvidia, line 323: did you get a perf increase across the board from that change, or what is the motivation for this? I feel like we should probably expose this to the specific calling site to make it nicely tunable.

mujoco/mjx/_src/solver.py
```diff
@@ -0,0 +1,710 @@
...
+  # ctx.search.zero_()
+  # ctx.gauss.zero_()
+  ctx.cost.fill_(wp.inf)
+  # ctx.prev_cost.zero_()
+  ctx.solver_niter.zero_()
+  # ctx.active.zero_()
+  ctx.done.zero_()
+
+  _update_constraint(m, d, ctx)
+  if wp.static(grad):
```
adenzler-nvidia, line 95: no reason to use wp.static outside of a kernel.

adenzler-nvidia (commented):
some comments - but got lost in the details, so I think I need to read the mujoco docs again first to have a better understanding of this.

From far, lots of interesting optimization opportunities, but I suggest to get this in rather quickly and focus on getting it correct first, and then let the profiler guide optimizations.

mujoco/mjx/_src/smooth.py
thowell: with `block_dim = 32` there is a runtime error that the cuda graph capture failed

mujoco/mjx/_src/solver.py
thowell: removed

mujoco/mjx/_src/smooth.py
adenzler-nvidia: I see, I think I'm hitting similar issues on the implicit integrator. What was the problematic MJCF? I think it will be tricky to make this tile API stuff bulletproof as a lot of this depends on the model..

erikfrey: I'm seeing this too fyi.  I know you guys fixed a tile API bug that was causing tests with pendula.xml to fail, and I was waiting til after that landed in nightly to try again.  Looks like it's still an issue.  Let me escalate this in chat.

mujoco/mjx/_src/io.py
```diff
@@ -141,14 +153,24 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   return m
 
 
-def make_data(mjm: mujoco.MjModel, nworld: int = 1) -> types.Data:
+def make_data(
+  mjm: mujoco.MjModel, nworld: int = 1, njmax: int = -1
+) -> types.Data:
   d = types.Data()
   d.nworld = nworld
+  d.nefc_total = wp.zeros((1), dtype=wp.int32, ndim=1)
```
erikfrey, line 161: nittiest nit in the whole world:

```py
wp.zeros((1,), ...
```

(needs a comma)

mujoco/mjx/_src/io.py
```diff
@@ -189,6 +211,14 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1) -> types.Data:
   d.qfrc_smooth = wp.zeros((nworld, mjm.nv), dtype=wp.float32)
   d.qfrc_constraint = wp.zeros((nworld, mjm.nv), dtype=wp.float32)
   d.qacc_smooth = wp.zeros((nworld, mjm.nv), dtype=wp.float32)
+  d.qfrc_constraint = wp.zeros((nworld, mjm.nv), dtype=wp.float32)
+  d.efc_J = wp.zeros((njmax, mjm.nv), dtype=wp.float32)
+  d.efc_D = wp.zeros((njmax), dtype=wp.float32)
```
erikfrey, line 216: same nit for these lines down: needs a comma if you meant it to be a tuple, otherwise remove parentheses

mujoco/mjx/_src/io.py
```diff
@@ -200,9 +230,21 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1) -> types.Data:
...
+  mjm: mujoco.MjModel, mjd: mujoco.MjData, nworld: int = 1, njmax: int = -1
+) -> types.Data:
   d = types.Data()
   d.nworld = nworld
+  d.nefc_total = wp.array([mjd.nefc * nworld], dtype=wp.int32, ndim=1)
+
+  if njmax == -1:
+    # TODO(team): heuristic for njmax
+    njmax = 512
+  d.njmax = njmax
```
erikfrey, line 243: on second thought, in C MuJoCo this field lives on Model - maybe it's better there?

mujoco/mjx/_src/solver.py
```diff
@@ -0,0 +1,755 @@
...
+
+
+def _linesearch(m: types.Model, d: types.Data, ctx: Context):
+  NV = m.nv
+
+  @wp.kernel
+  def _gtol(ctx: Context, m: types.Model):
+    worldid = wp.tid()
+    sum = float(0.0)
+    for i in range(NV):
```
erikfrey, line 344: i'm guessing this for loop deep down in the linesearch is one of the things to look at perf-wise

(no action for now)

mujoco/mjx/_src/solver.py
```diff
@@ -0,0 +1,755 @@
...
+  ctx.search = wp.empty(shape=(d.nworld, m.nv), dtype=wp.float32)
+  ctx.gauss = wp.empty(shape=(d.nworld), dtype=wp.float32)
+  ctx.cost = wp.empty(shape=(d.nworld), dtype=wp.float32)
+  ctx.prev_cost = wp.empty(shape=(d.nworld), dtype=wp.float32)
+  ctx.solver_niter = wp.empty(shape=(d.nworld), dtype=wp.int32)
+  ctx.active = wp.empty(shape=(d.njmax), dtype=wp.int32)
+  ctx.gtol = wp.empty(shape=(d.nworld), dtype=wp.float32)
+  ctx.mv = wp.empty(shape=(d.nworld, m.nv), dtype=wp.float32)
+  ctx.jv = wp.empty(shape=(d.njmax), dtype=wp.float32)
+  ctx.quad = wp.empty(shape=(d.njmax, 3), dtype=wp.float32)
```
erikfrey, line 51: i think those atomic expressions are expensive - perhaps we should make these dtype=wp.vec3 so that we only need one atomicAdd

mujoco/mjx/_src/solver.py
```diff
@@ -0,0 +1,755 @@
...
+        ls_ctx.lo.alpha[worldid] + ls_ctx.hi.alpha[worldid]
+      )
+
+    wp.launch(_alpha_lo_next_hi_next_mid, dim=(d.nworld), inputs=[ls_ctx])
+
+    _create_lspoint(ls_ctx.lo_next, m, d, ctx)
+    _create_lspoint(ls_ctx.hi_next, m, d, ctx)
+    _create_lspoint(ls_ctx.mid, m, d, ctx)
+
+    @wp.kernel
```
erikfrey, line 491: just a very general comment:

this looks like the MJX code, but my first instinct would have been to more closely hew to the C MuJoCo code.  I made some design decisions back in the day that were JAX-specific and may not apply here, and in particular the JAX code takes some shortcuts in the linesearch logic that might lead to slightly different numerical results.

it might be nice to try to make this look more like C MuJoCo (e.g. updateBracket, possibly record more explicit LSResult, etc) in a future pass

(no action for now)

erikfrey (approved):
Similar comment as @adenzler-nvidia - let's get this in but come back to it right away and start profiling and speeding up things up.  I left a couple comments on where I suspect perf opportunities lie.

mujoco/mjx/_src/smooth.py
adenzler-nvidia: I can try to repro today and see if I can find the issue - what is the exact repro, just 1 humanoid with the solve tests, and block size 32?

mujoco/mjx/_src/io.py
thowell: fixed

mujoco/mjx/_src/io.py
thowell: fixed

mujoco/mjx/_src/solver.py
thowell: done

mujoco/mjx/_src/io.py
thowell: added todos for moving `njmax` to `Model`

mujoco/mjx/_src/solver.py
thowell: kernels for `search` and `grad` now additionally compute dot products in order to avoid unnecessary for loops in other kernels.

PR #32, opened 2025-02-24
Implicit fast integrator

adenzler-nvidia:
Ready for review now

Implements implicit integration - only dense matrices for now. Some questions inline for stuff I'm unsure about.

Performance for 8k batches:
n_humanoids | MJX | MjWarp Dense 
1                    | 21M  | 83M
2                    | 3.7M | 19M
3                    | 1M    | 10M
4                    | 630k | 7.5M
5                    | 350k | 5.8M


mujoco/mjx/_src/forward.py
```diff
@@ -212,6 +214,174 @@ def add_damping_sum_qfrc_kernel_dense(m: Model, d: Data):
...
+        wp.tile_store(d.qfrc_integration[worldid], qfrc_combined, offset=offset_nv)
+
+      wp.launch_tiled(
+        qderiv_actuator_fused_kernel,
+        dim=(d.nworld, size),
+        inputs=[m, d, damping, adr],
+        block_dim=block_dim,
+      )
+
+    qderiv_tilesize_nv = m.qderiv_implicit_tilesize_nv.numpy()
```
erikfrey, line 353: thinking out loud: as these helper fields pile up we might want to think about how to organize them

no action for now

mujoco/mjx/_src/forward.py
```diff
@@ -212,6 +214,174 @@ def add_damping_sum_qfrc_kernel_dense(m: Model, d: Data):
...
+            shape=(tilesize_nv, tilesize_nv), dtype=wp.float32
+          )
+
+        if wp.static(passive_enabled):
+          dof_damping = wp.tile_load(damping, shape=tilesize_nv, offset=offset_nv)
+          negative = wp.tile_map(neg, dof_damping)
+          qderiv_tile = wp.tile_diag_add(qderiv_tile, negative)
+
+        # add to qM
+        qM_tile = wp.tile_load(
```
erikfrey, line 328: okay this is slick usage of tile api, very cool

mujoco/mjx/_src/io.py
```diff
@@ -188,6 +247,12 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   m.opt.timestep = wp.float32(mjm.opt.timestep)
   m.opt.impratio = wp.float32(mjm.opt.impratio)
 
+  # short-circuiting here allows us to skip a lot of code in implicit integration
+  m.actuator_affine_bias_gain = bool(
```
erikfrey, line 251: I try to avoid this - you can't always.  But what happens is that users - whether or not you tell them it's OK - change a biastype on the model and then wonder why the algorithm is failing.

What would you think of calculating this on the fly at the beginning of `implicit` instead of on the model?

mujoco/mjx/_src/types.py
```diff
@@ -163,6 +193,15 @@ class Model:
   qpos_spring: wp.array(dtype=wp.float32, ndim=1)
   body_tree: wp.array(dtype=wp.int32, ndim=1)  # warp only
   body_treeadr: wp.array(dtype=wp.int32, ndim=1)  # warp only
+  qderiv_implicit_offset_nv: wp.array(dtype=wp.int32, ndim=1)  # warp only
+  qderiv_implicit_offset_nu: wp.array(dtype=wp.int32, ndim=1)  # warp only
+  qderiv_implicit_tileadr: wp.array(dtype=wp.int32, ndim=1, device="cpu")  # warp only
```
erikfrey, line 198: that's interesting - does setting `device="cpu"` in the type spec actually do anything?

mujoco/mjx/_src/io.py
erikfrey: Well, arguing against myself, do we expect users to always use cudagraph with mjwarp functions if they want performance?  If we do, then it's OK to do a little bit of work host side during api calls.  What do you think?

mujoco/mjx/test_data/pendula_implicit.xml
```diff
@@ -0,0 +1,160 @@
+<!-- For validating dynamics of joints:
```
erikfrey, line 1: do we actually need to fork this file?  can you modify the original in such a way that it tests the stuff in smooth that we want, and it tests implicit features you want to test?

adenzler-nvidia (commented):
some question from my side that didn't make it because I forgot to push the "submit review" button

mujoco/mjx/_src/io.py
adenzler-nvidia: Performance will be very bad without graphs, so we definitely want things to be graphable. This both triggers a CPU side branch as well as a potential recompilation of the big fused kernel. Doing it at the start of implicit directly on GPU is certainly possible, but will most likely remove most of the performance benefits because now we need to compile the big version of the fused kernel all the time, which impact resource requirements and thus occupancy. Even if we branch inside the kernel and avoid all the unnecessary loads, we're not getting the speed of the direct early-out.

The tricky thing here is that "data is the API" so if a user changes this directly on GPU there is no way for us to figure out that it happened. That makes things very tricky. Which is an argument against doing this for sure. But then I guess anyone fiddling with the model directly can introduce lots of inconsistencies/bugs because many fields have dependencies coming from the compiler?

I tend to think of these things as "topology changes" vs "configuration changes". The first, we'll definitely need a recompilation (if you change things in the tree, add remove bodies, etc), the second one we shouldn't have to (for example changing a friction coefficient). But I'm not sure if that concept translates to MuJoCo directly, or whether there is something I'm missing. Is there a documentation somewhere about what users are allowed to change without recompiling?

So in the end - I'm unsure about what's best here. I think we have to see a bit how the performance of this looks in the big picture, and if we need similar techniques in other places. I do think we probably want them.

mujoco/mjx/_src/types.py
adenzler-nvidia: I'm pretty sure it doesn't. That's just a leftover from copy-pasting from io.py.

mujoco/mjx/_src/forward.py
adenzler-nvidia: it's starting to be difficult to get an overview, so I'm in favor of moving them out/grouping them better. And move the computation out into helper functions as well.

erikfrey: OK - I also wonder if some of these aren't actually OK to just calculate inline in the function where they are referenced, especially since we expect our users to use cudagraph.

WDYT?

(no action for now)

mujoco/mjx/_src/io.py
erikfrey: Yes your intuition is correct - discrete configurations like kinematic tree layout we expect a recompile, and continuous parameters like friction we don't.

Like my reply above, I would be OK with allowing a little more host-side computation of static helper variables such as this one happen during the relevant function call instead of during model creation.  We do this quite a bit in the current JAX code and it's fine.

mujoco/mjx/_src/forward.py
```diff
@@ -212,6 +214,174 @@ def add_damping_sum_qfrc_kernel_dense(m: Model, d: Data):
...
+  #
+  # The actuator_bias_gain_vel kernel could theoretically be fused in as well,
+  # but it's pretty clean straight-line code that loads a lot of data but
+  # only stores one array, so I think the benefit of keeping that one on-chip
+  # is likely not worth it compared to the compromises we're making with tile API.
+  # It would also need a different data layout for the biasprm/gainprm arrays
+  # to be tileable.
+
+  # assumptions
+  assert not m.opt.is_sparse  # unsupported
```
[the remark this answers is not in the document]
erikfrey: Ultimately we want a sparse option for everything, but this isn't a requirement for GTC.

I think what you have here is fine for GTC - we'll probably use dense for both demos.

mujoco/mjx/_src/io.py
```diff
@@ -114,6 +114,48 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     qLD_tileadr = np.cumsum(tile_off)[:-1]
     qLD_tilesize = np.array(sorted(tiles.keys()))
 
+  # tiles for implicit integration - needs nu + nv tile size and offset
+  qderiv_implicit_offset_nv = np.empty(shape=(0,), dtype=int)
```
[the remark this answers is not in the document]
erikfrey: If you buy my argument above that we calculate these inline in their respective functions, it becomes less of an issue.

mujoco/mjx/_src/io.py
```diff
@@ -188,6 +247,12 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   m.opt.timestep = wp.float32(mjm.opt.timestep)
   m.opt.impratio = wp.float32(mjm.opt.impratio)
 
+  # short-circuiting here allows us to skip a lot of code in implicit integration
+  m.actuator_affine_bias_gain = bool(
+    np.any(mjm.actuator_biastype == types.BiasType.AFFINE.value)
+    or np.any(mjm.actuator_gaintype == types.GainType.AFFINE.value)
+  )
```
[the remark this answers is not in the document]
erikfrey: Yes you are allowed and encouraged to do this.  Let's just try to put the computation of this variable in the function itself instead of as a Model field.

mujoco/mjx/_src/smooth.py
```diff
@@ -658,3 +658,41 @@ def solve_LD(m: Model, d: Data, L: array3df, D: array2df, x: array2df, y: array2
...
+    wp.launch_tiled(
+      cholesky, dim=(d.nworld, size), inputs=[m, adr, M, x, y], block_dim=block_dim
+    )
+
+  qLD_tileadr, qLD_tilesize = m.qLD_tileadr.numpy(), m.qLD_tilesize.numpy()
+
+  for i in range(len(qLD_tileadr)):
+    beg = qLD_tileadr[i]
+    end = m.qLD_tile.shape[0] if i == len(qLD_tileadr) - 1 else qLD_tileadr[i + 1]
+    tile_cholesky(beg, end - beg, int(qLD_tilesize[i]))
```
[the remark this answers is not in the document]
erikfrey: Factor and solve are split in the original code because we expose them as public API calls, as we try to stay close to the C MuJoCo API.

That said, fusing them makes a lot of sense, so we should keep both: factor/solve for public, fused for internal.

mujoco/mjx/_src/forward.py
adenzler-nvidia: I agree on just calculating inline, but I think for example this numpy() call would interfere with graph capture, as would any upload from CPU to GPU.

erikfrey: Oh really?  But when we created `qderiv_tilesize_nv` we explicitly set `device="cpu"` so surely `.numpy()` is doing nothing more here than returning a pointer or so?  I guess I'd have to try it to see.

adenzler-nvidia: The tilesizes + addresses are on CPU, but we need the dof offsets on GPU. I tried a bit and couldn't get it to work within a few tries, the other complicating thing is that you need other info in Model that is on GPU at that time, so we either have to maintain a CPU-side model as well or copy back.

The CPU-side model we kind of already have with the mujoco C model, but I'm not sure if that's ok because it could diverge from the model we use to simulate.

mujoco/mjx/_src/io.py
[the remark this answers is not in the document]
adenzler-nvidia: same problem here - the computation depends on data that resides on the GPU, so inlining it into the calling function will break graph capture.

mujoco/mjx/_src/forward.py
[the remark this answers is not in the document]
adenzler-nvidia: added a todo comment - let's do this after the initial crunch.

adenzler-nvidia:
merging this, I think the open points are all valid but can be done in follow-up PRs. We should definitely have a strategy about all the precomputation and compile-time constant generation.

PR #40, opened 2025-02-27
Collision Driver

adenzler-nvidia:
Sharing some of the collision driver WIP work to get feedback. The main interesting function is `collision` at the end of `collision_driver.py`

Here's our line of thought/how we think the sequence will look like:

1. broadphase: batched per-env, produces candidate pairs for narrow phase
2. filtering: filter out unwanted candidate pairs (parent-child, etc.)
3. narrow-phase preprocessing: go from per-world candidate pairs to per collision-geom type pairs across all worlds.
4. narrow-phase: 1 function per pair type, across all worlds
5. narrow-phase postprocessing: go back to per-world contact list, do the remainig work such as condim ordering etc.

UPDATE: mergable now, the description above is not really correct anymore. We are following the existing draft CUDA code quite closely now.


erikfrey:
Generally sounds OK but I wonder if you can get away with fewer sorts.  For example, do you really need step 5?  Since we map from a contact to some number of efc rows, is the world ordering important?

erikfrey:
I eyeballed the code and it looks like it's going in the right direction, no major concerns!  Excited to see this come together.

adenzler-nvidia:
For 4, we should definitely explore options to improve that. I have a few ideas but no good intuition about what will work, so those have to be explored after we have a first naive version. Statically checking is 1 option, megakernels is another one, etc. We'll see.

For 5, I'm sure I don't understand all the details yet. But efc rows will be per-world, right? We can do the per-world list building inside the contact generation as well, the big question there is whether the order of efc rows matters for determinism reasons, and whether determinism is a goal here.

erikfrey:
@adenzler-nvidia A contact is associated with a world but that doesn’t automatically mean it must be sorted by world.  Just that the contact has a wodid field.  Does that make sense?  If not we can hop on chat.

adenzler-nvidia:
it kind of makes sense - but further down when creating constraints out of contacts, you still want to group them by world? So having them sorted by world would mean that we have more efficient processing further down the line?

adenzler-nvidia:
update with a lot of changes - we aligned the flow much more with the already existing CUDA port of the pipeline as well as the earlier GJK warp port. Should be easy to integrate now.

I'm hope to iterator on the nits and bits in this state without adding more functionality to unblock everyone. Obvious missing things: collision filtering, actual implementation in the collision functions, and likely some missing fields in model/data.

mujoco/mjx/_src/types.py
```diff
@@ -342,11 +386,16 @@ class Data:
...
-  boxes_sorted: wp.array(dtype=wp.types.matrix(shape=(2, 3), dtype=wp.float32), ndim=2)
+  boxes_sorted: wp.array(dtype=wp.vec3, ndim=3)
   data_start: wp.array(dtype=wp.float32, ndim=2)
   data_end: wp.array(dtype=wp.float32, ndim=2)
   data_indexer: wp.array(dtype=wp.int32, ndim=2)
   ranges: wp.array(dtype=wp.int32, ndim=2)
   cumulative_sum: wp.array(dtype=wp.int32, ndim=1)
   segment_indices: wp.array(dtype=wp.int32, ndim=1)
-  geom_aabb: wp.array(dtype=wp.types.matrix(shape=(2, 3), dtype=wp.float32), ndim=1)
+  dyn_body_aamm: wp.array(dtype=wp.vec3, ndim=3)
```
eric-heiden, line 396: In the CUDA collision pipeline we use both `geom_aabb` and `geom_aamm`, but the AABBs can be indeed computed in a single kernel, so I would say we keep your change here (unlike what I mentioned in our discussion today).

eric-heiden, line 396: I think we need both `dyn_body_aamm` (shape `(nenv * nbody, 6)`) and `dyn_geom_aabb` (shape `(nenv * ngeom, 6)`). We can keep using 2x `wp.vec3` in the last dimension, or use `wp.spatial_vector` which also has `wp.vec3` accessors via `wp.spatial_top`, `wp.spatial_bottom`.

mujoco/mjx/_src/broad_phase_test.py
```diff
@@ -24,258 +24,14 @@
 
 from . import test_util
 
-BoxType = wp.types.matrix(shape=(2, 3), dtype=wp.float32)
```
adenzler-nvidia, line 27: This deletion is a mistake, these tests will come back for the final version.

mujoco/mjx/_src/collision_driver.py
```diff
@@ -382,7 +639,130 @@ def broad_phase(m: Model, d: Data) -> Data:
...
+    dim=(d.nworld, m.nbody),
+    inputs=[
+      m.body_geomnum,
+      m.body_geomadr,
+      m.geom_margin,
+      d.geom_xpos,
+      m.geom_rbound,
+      d.dyn_body_aamm,
+    ],
+  )
```
adenzler-nvidia, line 686: I think we are missing a step here where we collide geom<->geom, we're only doing the broadphase on bodies now. @nvtw

adenzler-nvidia: done.

mujoco/mjx/_src/collision_driver.py
```diff
@@ -208,93 +235,191 @@ def find_indices(
...
+  geom1 = geoms[0]
+  geom2 = geoms[1]
+
+  type1 = m.geom_type[geom1]
+  type2 = m.geom_type[geom2]
+  key = group_key(type1, type2)
+
+  n_type_pair = wp.atomic_add(d.narrowphase_candidate_group_count, key, 1)
+  d.narrowphase_candidate_worldid[key, n_type_pair] = worldid
+  d.narrowphase_candidate_geom[key, n_type_pair] = wp.vec2i(geom1, geom2)
```
erikfrey, line 407: Hmm OK I think I understand this - you're trading off a lot of memory for absolute certainty that when you launch a kernel for a certain narrowphase function, the candidates in that are correct.

At some point you may want to benchmark this against keeping all the candidates in a single array, and just early exiting if the types are wrong, e.g.:

```
@wp.kernel
def capsule_capsle(m, d):
  id1, id2 = wp.tid()
  type1, type2 = m.geom_type[id1], m.geom_type[id2]
  if type1 != capsule and type2 != capsule:
    return
```

No action for now, prefer to get this in as is.

mujoco/mjx/_src/collision_driver.py
```diff
@@ -208,93 +235,191 @@ def find_indices(
...
-    # world_id_j = j // num_boxes_per_world
-    j = j % num_boxes_per_world
+      id = wp.atomic_add(d.result_count, worldId, 1)
 
-    # assert worldId == world_id_j, "Only boxes in the same world can be compared"
-    # TODO: Remove print if debugging is done
-    # if worldId != world_id_j:
-    #     print("Only boxes in the same world can be compared")
+      if id < d.max_num_overlaps_per_world:
+        d.broadphase_pairs[worldId, id] = pair
```
Kenny-Vilella, line 329: I think that they are some inconsistency between bodies and geoms.
Here we store body in broadphase_pairs, but in the kernel group_contacts_by_type_kernel we consider this as geom.
This leads to an incorrect group_key associated with each contact.

Not sure whether we should reason with geom or body, but for sure we should be consistent throughout the pipeline.

mujoco/mjx/_src/io.py
```diff
@@ -316,9 +330,15 @@ def make_data(
...
   d.segment_indices = wp.array(segment_indices_list, dtype=int)
 
-  d.geom_aabb = wp.array(
-    mjm.geom_aabb, dtype=wp.types.matrix(shape=(2, 3), dtype=wp.float32), ndim=1
+  # internal narrowphase tmp arrays
+  ngroups = types.NUM_GEOM_TYPES
+  d.narrowphase_candidate_worldid = wp.empty(
+    (ngroups, d.ncon * nworld), dtype=wp.int32, ndim=2
+  )
+  d.narrowphase_candidate_geom = wp.empty(
```
erikfrey, line 338: It's important that the array shapes produced by `make_data` and `put_data` be the same.  So this shouldn't use `ncon` but that max overlaps variable instead.

erikfrey (approved):
Looks good, just nits.  My main comment is I'm not sure the (ngroups, d.max_num_overlaps_per_world * nworld) memory allocation is worth it just to avoid some early exiting, but let's get it in as is and then try it out.

mujoco/mjx/_src/support.py
```diff
@@ -154,3 +156,45 @@ def xfrc_accumulate(m: Model, d: Data) -> array2df:
   wp.launch(kernel=compute_qfrc, dim=(d.nworld, m.nv), inputs=[d, m, mask, qfrc_total])
 
   return qfrc_total
+
+
+@wp.func
```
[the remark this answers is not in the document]
eric-heiden: The problem with `wp.select` is the confusing ordering - the `if_false` argument comes before the `if_true` argument. It was meant to be used only internally, not really part of the public API. We will probably add a `wp.where` in the future with the argument ordering you expect from `torch` etc.

mujoco/mjx/_src/support.py
```diff
@@ -154,3 +156,45 @@ def xfrc_accumulate(m: Model, d: Data) -> array2df:
...
+  return c
+
+
+@wp.func
+def group_key(type1: wp.int32, type2: wp.int32) -> wp.int32:
+  return type1 + type2 * NUM_GEOM_TYPES
+
+
+@wp.func
+def all_same(a: wp.vec3, b: wp.vec3):
```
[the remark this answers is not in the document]
eric-heiden: That's right, I removed it now.

mujoco/mjx/_src/support.py
```diff
@@ -154,3 +156,45 @@ def xfrc_accumulate(m: Model, d: Data) -> array2df:
...
+  return a[0] == b[0] and a[1] == b[1] and a[2] == b[2]
+
+
+@wp.func
```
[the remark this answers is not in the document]
eric-heiden: Removed.

mujoco/mjx/_src/collision_functions.py
```diff
@@ -0,0 +1,192 @@
...
+def box_box(m: Model, d: Data, group_key: int):
+  pass
+
+
+def convex_convex(m: Model, d: Data, group_key: int):
+  pass
+
+
+@wp.func
+def group_key(type1: wp.int32, type2: wp.int32) -> wp.int32:
```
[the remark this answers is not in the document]
eric-heiden: Removed, using the `group_key` function from support now.

mujoco/mjx/_src/collision_driver.py
adenzler-nvidia: yes, this is a mistake. This should all be geom-centric. @nvtw could you take a look at this?

mujoco/mjx/_src/collision_driver.py
```diff
@@ -82,57 +118,51 @@ def overlap(
...
+  d_val = wp.dot(box_half_size, abs_dir)
+  f = center - d_val
 
   # Store results in the data arrays
-  data_start[worldId, i] = f
-  data_end[worldId, i] = center + d
-  data_indexer[worldId, i] = i
+  d.data_start[worldId, i] = f
+  d.data_end[worldId, i] = center + d_val
+  d.data_indexer[worldId, i] = i
```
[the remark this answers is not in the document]
nvtw: I renamed them to be more descriptive. the bounding boxes get projected onto the sweep direction. Each projection results in a range. data_start is the lower value of that range, data_end the upper value. data_indexer is uesed to track the reording during the sort. I renamed them to box_projections_lower, box_projections_upper and box_sorting_indexer

mujoco/mjx/_src/collision_driver.py
```diff
@@ -208,93 +235,191 @@ def find_indices(
...
+  mix = where((solmix1 < MJ_MINVAL) and (solmix2 >= MJ_MINVAL), 0.0, mix)
+  mix = where((solmix1 >= MJ_MINVAL) and (solmix2 < MJ_MINVAL), 1.0, mix)
+
+  p1 = m.geom_priority[g1]
+  p2 = m.geom_priority[g2]
+  mix = where(p1 == p2, mix, where(p1 > p2, 1.0, 0.0))
+  is_standard = (m.geom_solref[g1, 0] > 0) and (m.geom_solref[g2, 0] > 0)
+
+  solref_ = wp.vec(0.0, length=MJ_NREF, dtype=wp.float32)
+  for i in range(MJ_NREF):
```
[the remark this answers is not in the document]
nvtw: For the where call the loop is still required

mujoco/mjx/_src/collision_driver.py
```diff
@@ -208,93 +235,191 @@ def find_indices(
...
+    solref_[i] = where(
+      is_standard, solref_[i], wp.min(m.geom_solref[g1, i], m.geom_solref[g2, i])
+    )
 
-    # Check if the boxes overlap
-    if idx1 != idx2 and overlap(box1, boxes_sorted[worldId, j]):
-      pair = wp.vec2i(wp.min(idx1, idx2), wp.max(idx1, idx2))
+  friction_ = wp.vec3(0.0, 0.0, 0.0)
+  for i in range(3):
+    friction_[i] = wp.max(m.geom_friction[g1, i], m.geom_friction[g2, i])
```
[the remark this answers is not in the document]
nvtw: I'm also not sure here since m.geom_friction does not explicitly contain vec3. Maybe it works. @eric-heiden do you know if the loops can be avoided?

mujoco/mjx/_src/collision_driver.py
```diff
@@ -208,93 +235,191 @@ def find_indices(
...
-  direction = wp.vec3(0.5935, 0.7790, 0.1235)
-  direction = wp.normalize(direction)
-  abs_dir = wp.vec3(abs(direction.x), abs(direction.y), abs(direction.z))
+@wp.kernel
+def group_contacts_by_type_kernel(
+  m: Model,
+  d: Data,
+):
+  worldid, tid = wp.tid()
+  if tid >= d.result_count[worldid]:
```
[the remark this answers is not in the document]
nvtw: Changed it, result_count is indeed not a great name

mujoco/mjx/_src/collision_driver.py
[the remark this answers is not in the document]
eric-heiden: I converted the solver params to vectors now which eliminates the for loops.

mujoco/mjx/_src/collision_driver.py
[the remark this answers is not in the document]
eric-heiden: Yes, this is resolved now.

mujoco/mjx/_src/collision_driver.py
```diff
@@ -17,59 +17,95 @@
...
+  aabb_size = m.geom_aabb[gid, 1]
+
+  aabb = transform_aabb(aabb_pos, aabb_size, pos, ori)
+
+  # Write results to output
+  d.dyn_geom_aabb[env_id, gid, 0] = aabb.min
+  d.dyn_geom_aabb[env_id, gid, 1] = aabb.max
+
+
+@wp.kernel
```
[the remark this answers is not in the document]
eric-heiden: I think this kernel is indeed not necessary, we will only evaluate ncon `Contact` parameters which are already written by the following kernels. I removed it.

mujoco/mjx/_src/collision_driver.py
nvtw: It's fixed now

mujoco/mjx/_src/io.py
adenzler-nvidia: I will make another PR with a test that enforces this.

mujoco/mjx/_src/collision_driver.py
erikfrey: @adenzler-nvidia actually now that I see #53 from @thowell let's quickly think through what is a minimal set of arrays that accomodates both.

I suspect that something like option 3 might be OK for now as I'm not sure it's the long pole in the step function, but I could certainly be wrong.  During this early phase of development and especially as we have a lot of folks coordinating, I'd opt for simple + small perf hit.

Cross-posting from that PR, could you and @thowell work out a shared scheme of arrays that accomodates both broad phase methods?  If I were doing this myself, I would start by removing the contact type grouping dimension on candidates and just see what the perf looks like.  But I'll let you two sort out the details.

adenzler-nvidia: I'm not so sure about the "small perf hit" - doing some napkin math: let's have n broadphase overlaps.

For the current version, we have:
n memory accesses during group key sorting, atomic add, n memory accesses during narrowphase.

For the version with 1 array and early outs, we have
n * 32 memory accesses

I might be fine for now as we only have a few pair types implemented, but with all the combinations it becomes expensive quickly.

But in the end the grouping by type can be independent of the broadphase interface, it's just an additional step.

PR #43, opened 2025-02-28
modify linesearch point evaluation

thowell:


erikfrey:
@thowell Can you add a description?  I'll still real the code but it helps me to understand your thinking in this change.

erikfrey:
At first glance it looks like you're dropping some temp arrays and fusing a couple kernels - is that the basic gist?

erikfrey:
Going to merge this because I think I get it, but next time give me a little prose and poetry in the description       :-)

erikfrey (approved):

PR #45, opened 2025-03-01
Fewer deps faster tests.

erikfrey:


pyproject.toml
```diff
@@ -28,18 +28,14 @@ requires-python = ">=3.9"
 dependencies = [
     "absl-py",
     "etils[epath]",
-    "jax",
-    "jaxlib",
     "mujoco>=3.2.7",
     "pytest",
-    "scipy",
     "trimesh",
```
thowell, line 33: do we need trimesh?

thowell (approved):

pyproject.toml
erikfrey: We might need it for convex - if not I'll remove it after that code lands.

PR #46, opened 2025-03-03
Tiled mul_m.

erikfrey:
Result of `mjx-testspeed --function=solve --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192`

This branch:

```

 Total JIT time: 14.99 s
 Total simulation time: 1.33 s
 Total steps per second: 6,141,383
 Total realtime factor: 30,706.92 x
 Total time per step: 162.83 ns
```

main:

```

 Total JIT time: 0.89 s
 Total simulation time: 1.53 s
 Total steps per second: 5,363,853
 Total realtime factor: 26,819.26 x
 Total time per step: 186.43 ns
```

Note that compilation for tiled `mul_m` is quite slow:

```
Module mujoco.mjx._src.support 9a7a6e8 load on device 'cuda:0' took 14520.68 ms  (compiled)
```

Left a TODO - might need some help to debug what's going on there later.

### DIFF
--- a/mujoco/mjx/_src/support.py
+++ b/mujoco/mjx/_src/support.py
@@ -40,7 +40,7 @@ def mul_m(
     def tile_mul(adr: int, size: int, tilesize: int):
       # TODO(team): speed up kernel compile time (14s on 2023 Macbook Pro)
       @wp.kernel
-      def mul(m: Model, d: Data, leveladr: int, vec: array3df, res: array3df):
+      def mul(m: Model, d: Data, leveladr: int, res: array3df, vec: array3df):
         worldid, nodeid = wp.tid()
         dofid = m.qLD_tile[leveladr + nodeid]
         qM_tile = wp.tile_load(
--- a/mujoco/mjx/_src/support.py
+++ b/mujoco/mjx/_src/support.py
@@ -38,6 +38,7 @@ def mul_m(
   if not m.opt.is_sparse:
 
     def tile_mul(adr: int, size: int, tilesize: int):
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
Looks like I can only merge, but not approve.. Looks good! We should definitely have a look at compile times soon, it starts to get a bit out of hand, especially for the tile API kernels.

mujoco/mjx/_src/support.py
```diff
@@ -35,21 +36,42 @@ def mul_m(
...
+      @wp.kernel
+      def mul(m: Model, d: Data, leveladr: int, vec: array3df, res: array3df):
+        worldid, nodeid = wp.tid()
+        dofid = m.qLD_tile[leveladr + nodeid]
+        qM_tile = wp.tile_load(
+          d.qM[worldid], shape=(tilesize, tilesize), offset=(dofid, dofid)
+        )
+        vec_tile = wp.tile_load(vec[worldid], shape=(tilesize, 1), offset=(dofid, 0))
+        res_tile = wp.tile_zeros(shape=(tilesize, 1), dtype=wp.float32)
+        wp.tile_matmul(qM_tile, vec_tile, res_tile)
```
thowell, line 51: is there a performance benefit for the in-place function compared to `res_tile = wp.tile_matmul(qM_tile, vec_tile)`?

thowell (approved):

mujoco/mjx/_src/support.py
erikfrey: No noticeable perf impact but it's 1 less line of code   :-)     will get it into a future PR.

PR #48, opened 2025-03-03
modifications to solver

thowell:


erikfrey (approved):
This looks like mostly cleanup but I do see a few things that look like perf improvements, reducing number of kernel lanches etc.

Going to merge but don't forget to post benchmarks for any changes that touch perf!

PR #49, opened 2025-03-03
parallel linesearch

thowell:
this pr implements a parallel linesearch for the engine solver. the benchmarks below compare this parallel approach to the existing iterative linesearch.

the parallel linesearch will evaluate `ls_iterations` step sizes that are linearly spaced between zero and one and select the step size with the lowest cost. the iterative linesearch will sequentially evaluate step sizes until the maximum number of iterations `ls_iterations` is reached or the linesearch conditions are achieved.

**solver=cg, iterations=10, ls_iterations=20**

parallel:
```
mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=cg --iterations=10 --ls_itera
tions=20 --ls_parallel=True
```
```

 Total JIT time: 0.42 s
 Total simulation time: 8.18 s
 Total steps per second: 1,001,531
 Total realtime factor: 5,007.66 x
 Total time per step: 998.47 ns
```
 
iterative:
```
mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=cg --iterations=10 --ls_iterations=20 --ls_parallel=False
```
```

 Total JIT time: 0.63 s
 Total simulation time: 8.78 s
 Total steps per second: 932,921
 Total realtime factor: 4,664.60 x
 Total time per step: 1071.90 ns
```
 
 **solver=newton, iterations=10, ls_iterations=20**

parallel:
```
mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton --iterations=10 --ls_i
terations=20 --ls_parallel=True
```
```

 Total JIT time: 0.41 s
 Total simulation time: 23.64 s
 Total steps per second: 346,540
 Total realtime factor: 1,732.70 x
 Total time per step: 2885.67 ns
```
 
iterative:
```
 mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton --iterations=10 --ls_iterations=20 --ls_parallel=False
```
```

 Total JIT time: 0.62 s
 Total simulation time: 23.93 s
 Total steps per second: 342,297
 Total realtime factor: 1,711.49 x
 Total time per step: 2921.44 ns
```


erikfrey:
Just so it's clearer can you report benchmark numbers of just `solve`?

mujoco/mjx/_src/solver.py
```diff
@@ -4,6 +4,15 @@
 from . import support
 from . import types
 
+MAX_LS_PARALLEL = 64
```
erikfrey, line 7: Could we somehow make this a function of `ls_iterations`?

Consider `ls_iterations=4`, there's only so many points you can test in the iterative version right?  lo, mid, hi in 4 steps.  It's probably much less than 64 no?

mujoco/mjx/_src/solver.py
```diff
@@ -32,6 +41,9 @@ class Context:
   beta_num: wp.array(dtype=wp.float32, ndim=1)
   beta_den: wp.array(dtype=wp.float32, ndim=1)
   done: wp.array(dtype=wp.int32, ndim=1)
+  alpha_candidate: wp.array(dtype=wp.float32, ndim=1)
+  cost_candidate: wp.array(dtype=veclsf, ndim=1)
```
erikfrey, line 45: Actually this might be one where you benefit from making a proper 2d array.

`warp.tile_min` may do you much better than `warp.argmin`

mujoco/mjx/_src/solver.py
```diff
@@ -710,6 +791,21 @@ def solve(m: types.Model, d: types.Data):
...
+  if m.opt.ls_parallel:
+
+    @wp.kernel
+    def _alpha_candidate(ctx: Context, m: types.Model):
+      tid = wp.tid()
+
+      if tid >= m.opt.ls_iterations:
+        return
+
+      ctx.alpha_candidate[tid] = float(tid) / float(wp.max(m.opt.ls_iterations - 1, 1))
```
erikfrey, line 805: shouldn't the divisor here be MAX_LS_PARALLEL?

erikfrey (commented):
Starting off with a few basic questions.  I think this should def. be faster than iterative, but let's update the benchmarks with the stuff I just comitted and compare apples to apples, just benchmarking solve.

I suspect there's still some perf wins, made a few comments on stuff I think might help perf.

thowell:
pr description is updated with latest benchmarking results

erikfrey:
Hmmm, what do you think?  Presume that users will no longer tune `iterations` and `ls_iterations` in MJWarp, so there's a bit more pressure on `linesearch_iterative` - does the benefit of parallel linesearch become more apparent?

Said differently, why not crank `ls_iterations` up to 20?  But keep the early return.  I think that would be a fair comparison?

mujoco/mjx/_src/solver.py
thowell: `ls_iterations` sets the number of candidate step sizes to evaluate and `MAX_LS_PARALLEL` is the maximum number of candidate step sizes that can be evaluated per world.

mujoco/mjx/_src/solver.py
thowell: the denominator is updated to be `max(min(ls_iterations, MAX_LS_PARALLEL) - 1, 1)`

thowell:
@erikfrey the pr description is updated with new benchmarking results that compare the parallel linesearch to the iterative linesearch. the benchmarking results for both linesearch approaches include early returns throughout the engine solver (#77).

mujoco_warp/_src/types.py
```diff
@@ -275,6 +286,9 @@ class Constraint:
     hi_next_alpha: alpha for next high point          (nworld,)
     mid: loss at mid_alpha                            (nworld, 3)
     mid_alpha: midpoint between lo_alpha and hi_alpha (nworld,)
+    alpha_candidate: step sizes for linesearch        (MAX_LS_PARALLEL,)
```
erikfrey, line 289: here's a thought, what if in `Model` we did:

```
class Model:
...
  nlsp: int
...
  alpha_candidates: wp.array(dtype=wp.float32, ndim=1)
```

this has three advantages:

* we don't need a `MAX_LS_PARALLEL`
* you calculate `alpha_candidate` during `put_model` instead of the step
* you can avoid one extra early return check (you no longer have to check `if fooid < m.opt.ls_iterations` because `nlsp` tells you exactly the kernel dims

wdyt?

erikfrey (changes requested):
One design question

mujoco_warp/_src/types.py
thowell: - `alpha_candidate` is moved from `Data` to `Model` and initialized in `put_model`
- `MAX_LS_PARALLEL` is removed and is currently replaced with `ls_iterations`

instead of introducing `nlsp` the option `ls_iterations` could be utilized (the implementation is now currently utilizing `ls_iterations` for convenience). do we think it is important to introduce a new size variable?

erikfrey: Oh, yes that works.  We may want to eventually distinguish, e.g. there's a world where we actually do multiple iterations of parallel line search, but that's a future problem.

mujoco_warp/_src/io.py
```diff
@@ -311,7 +315,9 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   return m
 
 
-def _constraint(nv: int, nworld: int, njmax: int) -> types.Constraint:
+def _constraint(
+  nv: int, nworld: int, njmax: int, ls_iterations: int
```
erikfrey, line 319: nit: maybe simpler at this point to just pass `mjm: mujoco.MjModel` ?

mujoco_warp/_src/types.py
```diff
@@ -275,6 +277,9 @@ class Constraint:
...
```
erikfrey, line 280: need to update shape in docstring and remove `alpha_candidate`

erikfrey (approved):
Looks great, just a few remaining nits.

mujoco_warp/_src/io.py
```diff
@@ -208,6 +209,9 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   m.actuator_moment_tilesize_nu = wp.array(
     actuator_moment_tilesize_nu, dtype=wp.int32, ndim=1, device="cpu"
   )
+  m.alpha_candidate = wp.array(
+    np.array(np.linspace(0.0, 1.0, m.opt.ls_iterations)), dtype=wp.float32
```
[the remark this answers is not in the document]
thowell: fixed

mujoco_warp/_src/io.py
thowell: done

mujoco_warp/_src/types.py
thowell: fixed

mujoco_warp/_src/types.py
thowell: added `nlsp` to `Model`. `nlsp` is set in `put_model` to `ls_iterations`.

mujoco/mjx/_src/solver.py
thowell: this kernel has been removed

mujoco/mjx/_src/solver.py
thowell: `MAX_LS_PARALLEL` has been removed

mujoco/mjx/_src/solver.py
thowell: added a todo

PR #53, opened 2025-03-04
n^2 broadphase

thowell:
```
mjx-testspeed --function=nxn_broadphase --mjcf=humanoid/humanoid.xml --batch_size=8192
```
```

 Total JIT time: 0.01 s
 Total simulation time: 0.06 s
 Total steps per second: 140,677,877
 Total realtime factor: 703,389.38 x
 Total time per step: 7.11 ns
```

### DIFF
--- a/mujoco/mjx/_src/collision_driver.py
+++ b/mujoco/mjx/_src/collision_driver.py
@@ -523,8 +523,6 @@ def nxn_broadphase(m: Model, d: Data):
       + (m.ngeom - geom1) * ((m.ngeom - geom1) - 1) // 2
     )
 
-    geomtype1 = m.geom_type[geom1]
-    geomtype2 = m.geom_type[geom2]
     margin1 = m.geom_margin[geom1]
     margin2 = m.geom_margin[geom2]
     pos1 = d.geom_xpos[worldid, geom1]
--- a/mujoco/mjx/_src/collision_driver.py
+++ b/mujoco/mjx/_src/collision_driver.py
@@ -470,7 +470,7 @@ def broadphase_sweep_and_prune(m: Model, d: Data):
 def nxn_broadphase(m: Model, d: Data):
   filterparent = not (m.opt.disableflags & mujoco.mjtDisableBit.mjDSBL_FILTERPARENT)
 
-  d.nbroadphase_total.zero_()
+  d.broadphase_result_count.zero_()
 
   @wp.kernel
   def _nxn_broadphase(m: Model, d: Data):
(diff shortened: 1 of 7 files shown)


mujoco/mjx/_src/collision_driver.py
```diff
@@ -386,3 +387,79 @@ def broad_phase(m: Model, d: Data) -> Data:
...
+        wp.sqrt(float(-8 * elementid + 4 * m.ngeom * (m.ngeom - 1) - 7)) / 2.0 - 0.5
+      )
+    )
+    geom2 = (
+      elementid
+      + geom1
+      + 1
+      - m.ngeom * (m.ngeom - 1) // 2
+      + (m.ngeom - geom1) * ((m.ngeom - geom1) - 1) // 2
+    )
```
adenzler-nvidia, line 413: would be interesting to compare this to precomputing the index list and just fetching the two geoms to compare from memory. This is a lot of math for just an index, and sqrt does not have very good throughput on GPU.

My intuition could be totally wrong though.

erikfrey: Could be!  But at 7.19ns we're currently in "rounding error" territory perf wise compared to the rest of the step.  So I'm okay with doing something a bit dumb for now, can always come back to this later if it becomes signfiicant.

erikfrey:
The implementation LGTM, but can you work with @adenzler-nvidia to make sure that this algorithm populates the same fields as the other broadphase?  Here is what I would love to see:

* a single `broadphase` function that switches between nxn and sweep and prune based on `ngeom
  * (you'll want to do a bit of benchmarking to find a decent heuristic)
* a single set of broadphase result arrays that we then feed into narrowphase

mujoco/mjx/_src/collision_driver.py
```diff
@@ -386,3 +387,79 @@ def broad_phase(m: Model, d: Data) -> Data:
...
+    parent_child_collision = (
+      filterparent
+      and (weldid1 != 0)
+      and (weldid2 != 0)
+      and ((weldid1 == weld_parentid2) or (weldid2 == weld_parentid1))
+    )
+    mask = (contype1 and conaffinity2) or (contype2 and conaffinity1)
+
+    if (
+      (wp.norm_l2(pos2 - pos1) <= (size1 + size2))
```
erikfrey, line 442: can you double check if we also need to put margin in this calculation?

thowell: mjx is not using `margin` in the calculation https://github.com/google-deepmind/mujoco/blob/0f563ecf31bf178424fda4eb320e83c1de4f0cf7/mjx/mujoco/mjx/_src/collision_driver.py#L395

mujoco uses `margin` as part of the calculation in the narrowphase and midphase collision detector with a call to [mj_collideGeoms](https://github.com/google-deepmind/mujoco/blob/0f563ecf31bf178424fda4eb320e83c1de4f0cf7/src/engine/engine_collision_driver.c#L325) -> [mj_filterSphere](https://github.com/google-deepmind/mujoco/blob/0f563ecf31bf178424fda4eb320e83c1de4f0cf7/src/engine/engine_collision_driver.c#L146) 

will add margin to the calculation here

erikfrey: Yep, due to restrictions of static array sizes, MJX always outputs every contact every time.

adenzler-nvidia: I guess we need to add margin to the SAP broadphase as well? @nvtw

adenzler-nvidia:
I think on the broadphase output side we already have a pretty clean interface in the SAP - just 2 arrays, broadphase_result_count and broadphase_pairs. Both per-world. I would suggest to reuse those, and then defer the geomtype writeout and grouping to the group_contacts_by_type kernel.

I would also make a wp.func() for the filtering and call that from both the SAP and NxN broadphase, to make sure we're not duplicating that critical piece of code.

thowell:
@adenzler-nvidia 
- `nxn_broadphase` is updated to use `broadphase_result_count` and `broadphase_pairs`
- added two functions: 
  - `_geom_filter` for filtering geoms 
  - `_geom_pair` for setting the geom id order based on `geom_type`.

thowell:
@erikfrey `broadphase` is updated to switch between n^2 and sap based on ngeom

erikfrey (approved):
Looks excellent!  Just one nit then let's merge.

erikfrey:
Oh also can you do a `ruff format` ?

mujoco/mjx/_src/collision_driver.py
```diff
@@ -463,21 +460,65 @@ def broadphase_sweep_and_prune(m: Model, d: Data):
     inputs=[m, d, num_sweep_threads, filter_parent],
   )
 
+  return d
 
-###########################################################################################3
 
+def nxn_broadphase(m: Model, d: Data):
+  filterparent = not (m.opt.disableflags & mujoco.mjtDisableBit.mjDSBL_FILTERPARENT)
```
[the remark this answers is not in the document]
thowell: done

PR #54, opened 2025-03-05
Faster iterative linesearch

erikfrey:
Still a niggling numerical instability but this is mostly working, will fix broken tests tomorrow.

I suspect the things that contributed the most to the speedup are:

- fusing some kernel launches
- using fewer atomic adds
- implemented very simple early exiting

Results of `mjx-testspeed --function=solve --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192`

This PR:

```
 Total JIT time: 0.23 s
 Total simulation time: 1.07 s
 Total steps per second: 7,659,559
 Total realtime factor: 38,297.79 x
 Total time per step: 130.56 ns
```

Main:

```

 Total JIT time: 0.52 s
 Total simulation time: 1.27 s
 Total steps per second: 6,431,488
 Total realtime factor: 32,157.44 x
 Total time per step: 155.49 ns
```

Early exiting seems to be working quite nicely!   🎉    There is only a modest slowdown for 50 ls iterations with early exiting, compared to not early exiting.

Results of `mjx-testspeed --function=solve --is_sparse=False --mjcf=humanoid/humanoid.xml --batch_size=8192 --ls_iterations=20`

This PR:

```
 Total JIT time: 0.27 s
 Total simulation time: 1.14 s
 Total steps per second: 7,215,686
 Total realtime factor: 36,078.43 x
 Total time per step: 138.59 ns
```

Main:

```
 Total JIT time: 3.94 s
 Total simulation time: 4.30 s
 Total steps per second: 1,904,803
 Total realtime factor: 9,524.02 x
 Total time per step: 524.99 ns
```

### DIFF
--- a/mujoco/mjx/_src/solver.py
+++ b/mujoco/mjx/_src/solver.py
@@ -116,61 +116,6 @@ def _create_context(ctx: Context, m: types.Model, d: types.Data, grad: bool = Tr
     wp.launch(_search, dim=(d.nworld, m.nv), inputs=[ctx])
 
 
-def _eval_lspoint(ls_pnt: LSPoint, m: types.Model, d: types.Data, ctx: Context):
-  @wp.kernel
-  def _cost_deriv_gauss(ls_pnt: LSPoint, ctx: Context):
-    worldid = wp.tid()
-    alpha = ls_pnt.alpha[worldid]
-    alpha_sq = alpha * alpha
-    quad_total0 = ctx.quad_gauss[worldid][0]
-    quad_total1 = ctx.quad_gauss[worldid][1]
-    quad_total2 = ctx.quad_gauss[worldid][2]
-
-    ls_pnt.cost[worldid] = alpha_sq * quad_total2 + alpha * quad_total1 + quad_total0
-    ls_pnt.deriv_0[worldid] = 2.0 * alpha * quad_total2 + quad_total1
-    ls_pnt.deriv_1[worldid] = 2.0 * quad_total2
-
-  wp.launch(_cost_deriv_gauss, dim=(d.nworld,), inputs=[ls_pnt, ctx])
-
-  @wp.kernel
-  def _cost_deriv_quad(ls_pnt: LSPoint, ctx: Context, d: types.Data):
-    efcid = wp.tid()
-
-    if efcid >= d.nefc_total[0]:
-      return
-
-    worldid = d.efc_worldid[efcid]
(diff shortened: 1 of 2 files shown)


mujoco/mjx/_src/solver.py
```diff
@@ -361,65 +271,74 @@ def _rescale(m: types.Model, value: float) -> float:
...
-  wp.launch(_jv, dim=(d.njmax, m.nv), inputs=[ctx, d])
+  wp.launch(_jv, dim=(d.njmax, m.nv), inputs=[d, ctx])
 
   # prepare quadratics
   # quad_gauss = [gauss, search.T @ Ma - search.T @ qfrc_smooth, 0.5 * search.T @ mv]
   ctx.quad_gauss.zero_()
 
   @wp.kernel
-  def _quad_gauss(ctx: Context, m: types.Model, d: types.Data):
+  def _init_quad_gauss(ctx: Context, m: types.Model, d: types.Data):
```
thowell, line 328: nit: the other linesearch functions have `ctx: Context` as the final argument

mujoco/mjx/_src/solver.py
```diff
@@ -428,277 +347,303 @@ def _quad(ctx: Context, d: types.Data):
...
-  wp.launch(_quad, dim=(d.njmax), inputs=[ctx, d])
+    quad = wp.vec3()
+    quad[0] = 0.5 * Jaref * Jaref * efc_D
+    quad[1] = jv * Jaref * efc_D
+    quad[2] = 0.5 * jv * jv * efc_D
+    ctx.quad[efcid] = quad
+
+  wp.launch(_init_quad, dim=(d.njmax), inputs=[d, ctx])
+
+  # linesearch points
```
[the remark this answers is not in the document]
erikfrey: Hopefully not, here's my reasoning:

- I need to benchmark it but I suspect our use of malloc is so minimal right now that it won't make a difference perf-wise.
- From a usability perspective, we benefit from making the mjx/mjwarp structs look as much like C MuJoCo as possible.
- To get the JAX stuff working we may need to modify the kernels to take only arrays as inputs... if so I'd like to keep the function signatures as tight as possible or we are all gonna get carpal tunnel from writing these kernel function params     :-)

mujoco/mjx/_src/solver.py
erikfrey: Good catch thanks.

PR #55, opened 2025-03-05
hookup implicit integrator in step function

adenzler-nvidia:
one-line change, somehow forgot about this.


erikfrey:
@thowell has you covered in #51, going to merge that one instead just to keep it simple and avoid conflicts.

PR #57, opened 2025-03-05
WIP version of cylinder - cylinder collision

Kenny-Vilella:
Added a WIP version of cylinder - cylinder collision.
Still needs more testing.


erikfrey:
This might warrant some discussion - in MuJoCo we use the convex functions for cylinder<>cylinder.

We've tried some variants of this gradient descent approach in the past, and couldn't get a satisfactory perf for them compared to just running the convex algos on a decimated mesh that approximates the cylinder reasonably well.

I don't think we need cylinder<>cylinder for GTC, so I won't urgently review this PR but let me know if I'm missing something about needing this right away.

Then either here or in a video chat we can discuss what we want to do.

Kenny-Vilella:
@adenzler-nvidia Do we need this implemented for the GTC demo?

adenzler-nvidia:
Indeed it looks like we don't need it for G1 or Leap hand, but I'm sure there is a reason this is in the sheet. I just don't remember. But certainly a good thing to have, even if we don't have an immediate need for it.

erikfrey:
Let's get the GJK stuff sorted and then revisit this approach vs. automatically creating a mesh from the cylinder.  Maybe performance will be different in Warp than what we saw in JAX - it's definitely worth another look.

Going to close the PR for now but definitely want to come back to it probably not long after GTC.

PR #59, opened 2025-03-05
write_contact function that respects margin

adenzler-nvidia:
This should bring us the following benefits:
1. less contention on the atomic add of the total contact counter during contact generation
2. no more atomic add during constraint row setup.


nvtw:
I do support that change. It should not have any drawbacks except for maybe crazy high velocities which we should avoid anyway. And one can still increase the includeMargin parameter to make contacts appear a bit before they actually generate contact responses.

erikfrey (approved):

PR #61, opened 2025-03-05
Box-box collision

camevor:


erikfrey:
Hi @camevor - just checking, is this PR still WIP?  I would love to use this in leap hand training.

camevor:
Hi @erikfrey, yes, it's still marked WIP because I'm working on tests and it depends on two warp PR's that are not in yet (although fairly advanced in review). I'll try to get tests done and fix the bugs I find tomorrow, but if you'd like to try it out sooner, I can send you some patches.

erikfrey:
That timing works for me, thank you.

mujoco_warp/_src/collision_functions.py
```diff
@@ -289,6 +295,550 @@ def plane_capsule(
...
+  best_idx = wp.int32(face.best_idx)
+
+  if edge.best_dist < face.best_dist:
+    edge_axis, _ = get_box_axis(wp.int32(edge.best_idx), R)
+    if wp.abs(wp.dot(face_axis, edge_axis)) < 0.99:
+      best_axis = edge_axis
+      best_sign = wp.int32(edge.best_sign)
+      best_idx = wp.int32(edge.best_idx)
+  return best_axis, best_sign, best_idx
+  return wp.vec3(0.0), 1, 0
```
adenzler-nvidia, line 509: double return here.

mujoco_warp/_src/collision_functions.py
```diff
@@ -395,3 +945,5 @@ def narrowphase(m: Model, d: Data):
 
   for collision_kernel in _collision_kernels.values():
     wp.launch(collision_kernel, dim=d.nconmax, inputs=[m, d])
+  box_box(m, d)
```
adenzler-nvidia, line 948: I guess we need to go back discussing earlier changes a bit here - because with @thowell 's latest change we would have all the collision functions be a wp.func. That's fundamentally incompatible with the approach here that uses more than 1 thread per collision pair - and also means we cannot use tile API.

Questions:
1. is there a simpler version that could fit the wp.func interface that would also suffice, and could fit the generic interface?
2. maybe it makes sense to have the ones that fit the simple interface in a single launch, and have some specializations?

adenzler-nvidia (commented):
a few comments, we should discuss the launch architecture for sure.

Kenny-Vilella:
Some perf number I got using a scene with, 4 stacks of 5 boxes, 1024envs, since I got of out of memory with mjx if using more envs, and using the cg solver as the newton solver crash:
mjwarp:
```
 Total JIT time: 0.32 s
 Total simulation time: 1.01 s
 Total steps per second: 1,013,108
 Total realtime factor: 5,065.54 x
 Total time per step: 987.06 ns
```

mjx:
```
 Total JIT time: 6.70 s
 Total simulation time: 60.18 s
 Total steps per second: 17,015
 Total realtime factor: 85.08 x
 Total time per step: 58770.65 ns
```

When running the benchmark with the leap hand:
mjwarp:
```
 Total JIT time: 0.82 s
 Total simulation time: 6.91 s
 Total steps per second: 148,091
 Total realtime factor: 1,480.91 x
 Total time per step: 6752.59 ns
```

mjx:
```
 Total JIT time: 12.28 s
 Total simulation time: 1.32 s
 Total steps per second: 775,295
 Total realtime factor: 7,752.95 x
 Total time per step: 1289.83 ns
```

Will investigate a bit the leap hand scenario.
My guess is that it is either due to different dynamics influencing the results or how we filter contact.

Kenny-Vilella:
After investigation, the leap hand benchmark is slow because the leap hand explodes.
Even removing collision leads to explosion, although it is much later in the benchmark as collision seems to exacerbate the issue.
So my guess is that there are bugs somewhere else in the pipeline.
The limit constraint seems to be a good suspect.

erikfrey:
Woah, good catch!  Let's fix that - PR incoming.

erikfrey:
@Kenny-Vilella under what conditions are you seeing the explosion?

One weird thing about `mjwarp-testspeed` is that it overrides `iterations` and `ls_iterations` even if you don't specify params.  We will fix this soon.

So make sure you are setting `--iterations=5` and `--ls_iterations=8` to match the XML.

Kenny-Vilella:
I'm looking at the viewer actually.
Will it be also a problem in the viewer?
mjx behaves properly.

I'am using a leap hand xml from Alain.
What I see is even if I disable all the collision and constraint functions there is one of the finger that is oscillating for no reason.
And it seems to come from the Euler solver.
The implicit version seems to behave properly.

Kenny-Vilella:
I introduce in my local branch a hack to disable the solver step in case of no constraint.
If I adjust the contact parameter I can have a proper simulation without explosion and where the contact work correctly.
At a given set of parameter, we have huge diff between mjx and mjwarp, but I do not think it comes from the change in this PR.

Anyway, the contact generation looks fine and the perf is good when looking at scene where mjx and mjwarp behaves similarly, that is stack of box.

PR #63, opened 2025-03-06
Improve compilation times

eric-heiden:
This PR tries to reduce the kernel compilation times in two ways:

1. The kernels inside functions are moved at the beginning before any `wp.launch` calls happen. This ensures that we populate the module with all the kernels that are called later on. The first time you launch a kernel it will compile the module that contains this kernel. When we later declare more kernels via `@wp.kernel` they will be added to the same module (the Python file by default), and when we launch them we again may have to recompile the entire module. This is why you see compilation times for the same module (in this case `mujoco.mjx._src.solver`) to continuously increase, for example:
    ```
    Module mujoco.mjx._src.solver fdc3979 load on device 'cuda:0' took 436.63 ms  (compiled)
    Module mujoco.mjx._src.solver 643148f load on device 'cuda:0' took 496.40 ms  (compiled)
    Module mujoco.mjx._src.solver c0d5483 load on device 'cuda:0' took 806.94 ms  (compiled)
    Module mujoco.mjx._src.solver 1d909dc load on device 'cuda:0' took 826.02 ms  (compiled)
    Module mujoco.mjx._src.solver d12a861 load on device 'cuda:0' took 964.80 ms  (compiled)
    ```
   This step reduces the compilation time from 96.8s to 42.5s on the example below. I only fixed the `smooth`, `solver`, `support` and `forward` files for now, will do the remainder as well if this PR looks good.

2. I try to resolve another issue when we call functions from the same module. Consider the following example where we call 3 functions from the `forward` module:
    ```py
    fwd_position(m, d)
    fwd_velocity(m, d)
    fwd_actuation(m, d)
    ```
    The kernel launches inside each function trigger compilation of the `forward` module again (like before, the module corresponds to the Python file name by default). To avoid these successively larger recompilations of the same module I create a separate module inside the functions from the `smooth`, `solver`, `support` and `forward` modules. I added a custom `@kernel` decorator (we might want to add this directly to Warp's `@wp.kernel` decorator) which allows us to pass a `wp.context.Module` to the `Kernel` constructor so that we can programmatically create a Module which has a unique name (we can insert the tile sizes and other parameters also to the name string):
    ```py
    module = wp.context.Module(f"smooth._factor_solve_i_dense_{tilesize}", loader=None)

    @kernel(module=module)
        def cholesky(m: Model, leveladr: int, M: array3df, x: array2df, y: array2df):
            ...
    ```
    With this fix only the kernels and functions that are called within the Python functions like `fwd_position` are compiled to their own modules. This brought down the compilation time further to 18.4s for the example below. Set `USE_SEPARATE_MODULES` to False in `mjx/_src.__init__.py` to use the previous behavior with the module being the Python file.

## Example to compare compilation times:

```py
import mujoco
import mujoco.mjx
import numpy as np
from etils import epath
from mujoco.mjx._src.io import put_data, put_model

import warp as wp

wp.clear_kernel_cache()


def main():
    mjcf_filename = (epath.resource_path("mujoco.mjx") / "test_data/pendula.xml").as_posix()

    mjm = mujoco.MjModel.from_xml_path(mjcf_filename)
    dt = 0.01
    mjm.opt.timestep = dt
    mjm.opt.iterations = 1
    mjm.opt.ls_iterations = 4
    is_sparse = True
    mjm.opt.jacobian = is_sparse
    mjm.opt.solver = mujoco.mjtSolver.mjSOL_CG
    # mjm.opt.solver = mujoco.mjtSolver.mjSOL_NEWTON
    mjd = mujoco.MjData(mjm)
    mjd.qpos[1] = 0.1  # initial condition
    mujoco.mj_step(mjm, mjd, 3)  # let dynamics get state significantly non-zero
    mujoco.mj_forward(mjm, mjd)
    m = put_model(mjm)
    d = put_data(mjm, mjd)

    with wp.ScopedTimer("warmup"):
        mujoco.mjx.step(m, d)

if __name__ == "__main__":
    main()
```
Before this PR:
```
warmup took 96852.40 ms
```

After this PR:
```
warmup took 18478.45 ms
```


erikfrey:
Hi Eric - this is amazing, thank you.  I'm a bit out of my element reviewing this as I think it touches a bit on how Warp wants to think about things like modules, so I will leave a few general comments but I kind of want to hear from @adenzler-nvidia and possibly even Miles if this is sanctioned behavior          :-)

mujoco/mjx/_src/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+# ==============================================================================
+
+import functools
+import inspect
+from typing import Callable, Optional
+
+from warp.codegen import make_full_qualified_name
+from warp.context import Function, Kernel, Module, get_module
+
+USE_SEPARATE_MODULES = True
```
erikfrey, line 23: if we all agree this is a good practice, do we need this flag?  Why not just always use this new behavior?

mujoco/mjx/_src/forward.py
```diff
@@ -42,9 +44,14 @@ def _advance(
 ):
   """Advance state and time given activation derivatives and acceleration."""
 
+  if USE_SEPARATE_MODULES:
+    module = wp.context.Module("forward._advance", loader=None)
```
erikfrey, line 48: Could we play tricks with the `inspect` module so that we can generate this string automatically?  It would be nice if there was a way to keep this module string generation centralized and not distribute it throughout the code.

mujoco/mjx/_src/forward.py
```diff
@@ -281,7 +290,15 @@ def subtract_multiply(x: wp.float32, y: wp.float32):
     def qderiv_actuator_damping_tiled(
       adr: int, size: int, tilesize_nv: int, tilesize_nu: int
     ):
-      @wp.kernel
+      if USE_SEPARATE_MODULES:
+        module = wp.context.Module(
+          f"forward.qderiv_actuator_damping_tiled_{tilesize_nu}_{tilesize_nv}",
```
erikfrey, line 295: I wonder if we could kill two birds with one stone here:

In Jax, the `@jax.jit` decorator (which is kind of like `@wp.kernel`) has a neat parameter, `static_argnums`

It lets you tell Jax right away which params to the function should be treated as static.  If we supported this feature in your custom kernel parameter, we get two benefits:

- you have the info you need to strictly define your module name
- we can have one less nested function (I think) which makes the code cleaner

I think it would be possible then for kernels to look like this:

```
@kernel(static_argnums=(2,3))
def my_kernel(m: Model, d: Data, tilesize_nu: int, tilesize_nv: int)
  ...
```

WDYT?

erikfrey (commented):
Just a few general questions, but I'm looking mostly for feedback from NVIDIA folks.

mujoco/mjx/_src/__init__.py
eric-heiden: Yes, I will remove this flag and clean up the PR a bit further. It was just for testing the impact of separating out the kernels into separate modules.

mujoco/mjx/_src/forward.py
eric-heiden: I definitely like the idea to have these static args. I'm currently experimenting with a `@wp.kernel(module="unique")` setting that will generate unique module names automatically based on the kernel code and `wp.static`, closure variables, etc. (what the `ModuleHasher` is already doing anyway). So for this purpose I don't think we need it, other than having potentially less nested functions. But I'm not sure if we need this right now, it will take a bit more design to get such a feature implemented.

eric-heiden:
The new `@kernel` implementation relies on an updated version of Warp's `@wp.kernel` decorator which should be available in tomorrow's Warp nightly PIP package.

adenzler-nvidia:
I like it - we just need to make sure this only lands once the necessary changes are part of warp nightly.

Would also be great to document this in the readme or in our new best practices guide such that everyone can follow the pattern.

mujoco/mjx/_src/__init__.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2025 The Physics-Next Project Developers
```
erikfrey, line 1: In #67 I propose creating a file `warp_util.py` - do you think we could put this code in there?

I vaguely recall some rule of thumb that `__init__.py` should be reserved for imports to define the api of the module - I don't remember the details but I ususally stick to it if possible

erikfrey (changes requested):
OK this is awesome, just one nit/request, otherwise looks good to go from me.

mujoco/mjx/_src/__init__.py
eric-heiden: Yes, I just moved the `@kernel` decorator to `warp_util.py`.

erikfrey (approved):
:tada: so nice!

erikfrey:
I'm going to merge this because it's instant benefit, we can follow up with a separate PR to annotate everything else, but it looks like this PR already has a lot of it.

PR #64, opened 2025-03-06
Avoid compiling tile functions with 0-length tiles (avoids compiler errors)

adenzler-nvidia:


mujoco/mjx/_src/forward.py
```diff
@@ -349,9 +349,10 @@ def qderiv_actuator_fused_kernel(
       end = (
         m.qLD_tile.shape[0] if i == len(qderiv_tileadr) - 1 else qderiv_tileadr[i + 1]
       )
-      qderiv_actuator_damping_tiled(
-        beg, end - beg, int(qderiv_tilesize_nv[i]), int(qderiv_tilesize_nu[i])
-      )
+      if qderiv_tilesize_nv[i] != 0:
+        qderiv_actuator_damping_tiled(
```
adenzler-nvidia, line 353: only check nv here and do a wp.static for nu, because we need to run parts of the kernel even without actuators.

thowell (approved):
lgtm

PR #67, opened 2025-03-07
Trace option for testspeed and some updated benchmarks.

erikfrey:
With this PR we can get some detailed perf numbers of the subcomponents of the physics step.

I've updated the benchmark to compare the results of this to MJX on `humanoid.xml`.


adenzler-nvidia:
looks good from my side - good to have this critical piece of infrastructure!

PR #68, opened 2025-03-07
Tiled version of euler integration

adenzler-nvidia:
Pretty straightforward, similar to implicit integrator. There's more to be gained by fusing in advance as well but I need to figure out interface considerations/composability first.

Timings are pretty similar for a single humanoid, but for 5 humanoids there is a good speedup:

before:
```

 Total JIT time: 1.26 s
 Total simulation time: 12.44 s
 Total steps per second: 658,355
 Total realtime factor: 3,291.78 x
 Total time per step: 1518.94 ns
```

after:
```

 Total JIT time: 1.28 s
 Total simulation time: 11.32 s
 Total steps per second: 723,993
 Total realtime factor: 3,619.96 x
 Total time per step: 1381.23 ns
```

### DIFF
--- a/mujoco/mjx/_src/forward.py
+++ b/mujoco/mjx/_src/forward.py
@@ -163,7 +163,8 @@ def euler(m: Model, d: Data):
   """Euler integrator, semi-implicit in velocity."""
   # integrate damping implicitly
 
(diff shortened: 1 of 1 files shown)


erikfrey (approved):
Looks great.

PR #69, opened 2025-03-07
Capsule-capsule collision

jprozorova:


mujoco/mjx/_src/math_test.py
```diff
@@ -0,0 +1,185 @@
...
+  def test_overlapping_segments(self):
+    """Tests that perfectly overlapping segments intersect at the midpoints."""
+    a0, a1 = wp.vec3([0.0, 0.0, -1.0]), wp.vec3([0.0, 0.0, 1.0])
+    b0, b1 = wp.vec3([0.0, 0.0, -1.0]), wp.vec3([0.0, 0.0, 1.0])
+
+    best_a, best_b = closest_segment_to_segment_points(a0, a1, b0, b1)
+    self.assertSequenceAlmostEqual(best_a, [0.0, 0.0, 0.0], 5)
+    self.assertSequenceAlmostEqual(best_b, [0.0, 0.0, 0.0], 5)
+
+  params = list(zip(np.repeat(np.arange(2), 2), np.tile(np.arange(2), 2)))
```
erikfrey, line 167: Ah, I see you got this from the MJX tests.  I know the person who wrote this test - they are very rigorous about their testing and was probably using this test to verify the algorithm while building it.  Geometry code is tricky and hard to get right.

That's great, but I personally would have removed this test after validating.  Here's why:

* The previous tests arguably cover any "edge case" scenario in the algorithm
* The previous tests are more legible - if they fail, you have an explicit clue as to the cause of the failure
  * (e.g. oh it fails with zero length segments, maybe a divide by zero?)
* Tests with random parameters are prone to flakiness (not always!)
* Your validation code (including a fancy minimization algo) is arguably longer / more complex than the code you are testing.  If this test fails, imo the error is just as likely to be in the validation code than in the implementation!

Anyway this is somewhat on the border of personal preference - I would save a test like this for integration testing and not unit testing.  I'm going to leave this as an optional comment, if you personally see the value of the test then by all means leave it in.

erikfrey (approved):
This looks really nice - I left one comment but it's optional.

I'd like @adenzler-nvidia to take a look though as he's been coordinating the collision stuff.

mujoco/mjx/_src/math_test.py
jprozorova: makes sense,  left this test out, thanks!

adenzler-nvidia:
not much to add - nice and clean code. Thank you for formatting some of the stuff, we (I guess mostly myself) have not been very diligent about running the formatter because CI has been red anyway.

Could you run the formatter again on this? Then we can merge.

adenzler-nvidia:
merging this as it didn't break CI but we'll get the formatting right again - then we can focus on the test failure.

PR #70, opened 2025-03-07
Fix indentation and ruff error.

erikfrey:


adenzler-nvidia (approved):
oh wow - thanks for catching that. looks like there is no test hitting that line of code, something for me next week.

PR #71, opened 2025-03-07
fix put_data ncon initialization

thowell:


mujoco/mjx/_src/io.py
```diff
@@ -454,7 +454,7 @@ def put_data(
   if nworld * mjd.nefc > njmax:
     raise ValueError("nworld * nefc > njmax")
 
-  d.ncon = wp.array([mjd.ncon], dtype=wp.int32, ndim=1)
+  d.ncon = wp.array([mjd.ncon * nworld], dtype=wp.int32, ndim=1)
```
erikfrey, line 457: Oh snap good catch.

We'll eventually add some more tests but for now I'll just submit this.

erikfrey (approved):

PR #74, opened 2025-03-09
Fixes to collision driver.

erikfrey:
* Use plane<>bounds dist for plane colliders in broadphase
* Fix capsule<>plane generated normal (was transposed)
* Don't pass large strings to `test_contact` params (they get printed to stdout as part of the test name)
* Remove plane<>box test since we don't yet have that function
* Add plane<>capsule test
* Set condim for discovered contacts


mujoco/mjx/_src/collision_driver.py
```diff
@@ -490,16 +496,30 @@ def _nxn_broadphase(m: Model, d: Data):
...
+    xmat1 = d.geom_xmat[worldid, geom1]
+    xmat2 = d.geom_xmat[worldid, geom2]
     size1 = m.geom_rbound[geom1]
     size2 = m.geom_rbound[geom2]
 
     bound = size1 + size2 + wp.max(margin1, margin2)
     dif = pos2 - pos1
-    sphere_filter = wp.dot(dif, dif) <= bound * bound
+
+    if size1 != 0.0 and size2 != 0.0:
```
adenzler-nvidia, line 507: thanks for the comment - not knowing MuJoCo all that well yet it looks not straightforward to deduce geomtype from size like this.

But we also don't want to pay the price for a type lookup for sure here, so I guess the comment is as good as it gets.

No action required.

mujoco/mjx/_src/collision_functions.py
```diff
@@ -279,7 +278,8 @@ def plane_capsule(
     else:
       b = wp.vec3(0.0, 0.0, 1.0)
 
-  frame = mat33_from_cols(n, b, wp.cross(n, b))
+  c = wp.cross(n, b)
+  frame = wp.mat33(n[0], n[1], n[2], b[0], b[1], b[2], c[0], c[1], c[2])
```
adenzler-nvidia, line 282: there should be an equivalent mat33_from_rows as far as I remember.

adenzler-nvidia (approved):
Looks good, thank you!

adenzler-nvidia:
you might want to reinstate the box-plane test because that has been merged now.

mujoco/mjx/_src/collision_driver_test.py
```diff
@@ -17,25 +17,40 @@
...
 from mujoco import mjx
 import numpy as np
+from absl.testing import absltest
 from absl.testing import parameterized
 
 
-class ConvexTest(parameterized.TestCase):
-  """Tests the convex contact functions."""
+class PrimitiveTest(parameterized.TestCase):
+  """Tests the primtive contact functions."""
```
thowell, line 25: primitive

thowell (approved):

mujoco/mjx/_src/collision_functions.py
erikfrey: You know I tried it but I think I got an error too?  Not sure.  I'll come back and try it again later.

mujoco/mjx/_src/collision_driver.py
erikfrey: Yeah agreed it's a little weird, I'll actually make the comment more explicit about it.  Thanks for the note.

mujoco/mjx/_src/collision_driver_test.py
erikfrey: Fixed!  Thanks.

erikfrey:
@adenzler-nvidia OK added back box<>plane test but then got errors!  I noticed a few issues with box<>plane such as expecting size instead of half size (I've made this mistake before).... I started pulling on the thread but just to be safe, and in the interest of time, I made the algorithm 1:1 to MuJoCo's.  Submitting now.

PR #75, opened 2025-03-09
MJWarp viewer and get_data_into.

erikfrey:
* Adds an interactive viewer that calls MJWarp step
* Removes a `visualize.py` that should be covered by this new viewer
* Adds `get_data_into` which is needed by the viewer
* Removes contype/conaffinity/pair from `humanoid.xml` - not needed by MJWarp
* Raises iterations / ls_iterations in `humanoid.xml` since MJWarp has early exiting


mujoco/mjx/_src/io.py
```diff
@@ -658,3 +658,109 @@ def tile(x):
...
+        j = mjm.dof_parentid[j]
+        adr += 1
+    # TODO(team): set efc_J after fix to _realloc_con_efc lands
+    # if nefc > 0:
+    #   result.efc_J[:nefc * mjm.nv] = d.efc_J.numpy()[:nefc].flatten()
+  result.efc_D[:] = d.efc_D.numpy()[:nefc]
+  result.efc_pos[:] = d.efc_pos.numpy()[:nefc]
+  result.efc_aref[:] = d.efc_aref.numpy()[:nefc]
+  result.efc_force[:] = d.efc_force.numpy()[:nefc]
+  result.efc_margin[:] = d.efc_margin.numpy()[:nefc]
```
adenzler-nvidia, line 764: does this need a UT like the make_data/put_data functions to make sure we're not missing things?

No action requred, can be added later.

adenzler-nvidia (approved):
Thank you very much!

PR #76, opened 2025-03-10
Bugfix damping index in add_damping_sum_qfrc_kernel_sparse

Kenny-Vilella:
Fixed an index issue in add_damping_sum_qfrc_kernel_sparse

### DIFF
--- a/mujoco/mjx/_src/forward.py
+++ b/mujoco/mjx/_src/forward.py
@@ -173,10 +173,7 @@ def euler(m: Model, d: Data):
       worldId, tid = wp.tid()
 
       dof_Madr = m.dof_Madr[tid]
-      i = dof_Madr // m.nv
-      j = dof_Madr % m.nv
-      damping = wp.select(i == j, 0.0, m.opt.timestep * m.dof_damping[i])
-      d.qM_integration[worldId, 0, dof_Madr] += damping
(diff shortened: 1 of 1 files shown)


mujoco/mjx/_src/forward.py
```diff
@@ -173,7 +173,10 @@ def add_damping_sum_qfrc_kernel_sparse(m: Model, d: Data):
       worldId, tid = wp.tid()
 
       dof_Madr = m.dof_Madr[tid]
-      d.qM_integration[worldId, 0, dof_Madr] += m.opt.timestep * m.dof_damping[dof_Madr]
+      i = dof_Madr // m.nv
+      j = dof_Madr % m.nv
+      damping = wp.select(i == j, 0.0, m.opt.timestep * m.dof_damping[i])
+      d.qM_integration[worldId, 0, dof_Madr] += damping
```
adenzler-nvidia, line 179: I think you can just call dof_damping[tid] here - we're launching nv threads per world.

Kenny-Vilella: Thanks !
Didn't notice that Madr is the diagonal element.
It is now fixed.

adenzler-nvidia:
Thanks for spotting this - looks good now!

PR #77, opened 2025-03-10
solver early return

thowell:
this pr aims to improve the performance of the engine solver by incorporating early returns in kernels launched by `solve`.

the first benchmarking results evaluate overhead introduced by the early returns by considering one solver iteration:

**solver=cg, iterations=1, ls_iterations=20**
```
mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=cg --iterations=1 --ls_iterations=20
```

this pr:
```

 Total JIT time: 4.36 s
 Total simulation time: 3.98 s
 Total steps per second: 2,058,020
 Total realtime factor: 10,290.10 x
 Total time per step: 485.90 ns
```

main:
```

 Total JIT time: 0.10 s
 Total simulation time: 3.97 s
 Total steps per second: 2,065,483
 Total realtime factor: 10,327.42 x
 Total time per step: 484.15 ns
 ```
 

 **solver=newton, iterations=1, ls_iterations=20**
 ```
 mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton --iterations=1 --ls_iterations=20
 ```
 
 this pr:
 ```

 Total JIT time: 2.78 s
 Total simulation time: 8.01 s
 Total steps per second: 1,022,150
 Total realtime factor: 5,110.75 x
 Total time per step: 978.33 ns
 ```
 
 main:
 ```
 mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton --iterations=1 --ls_iterations=20
 ```
 ```

 Total JIT time: 0.10 s
 Total simulation time: 7.99 s
 Total steps per second: 1,024,950
 Total realtime factor: 5,124.75 x
 Total time per step: 975.66 ns
 ```
 
 **for both cg and newton solvers a small amount of overhead is introduced by the early returns**
 
 the second set of benchmarks evaluate the effect of early returns on the engine solver performance by considering multiple solver iterations:
 
 **solver=cg, iterations=10, ls_iterations=20**
 ```
 mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=cg --iterations=10 --ls_iterations=20
 ```
 
 this pr:
 ```

 Total JIT time: 4.95 s
 Total simulation time: 8.68 s
 Total steps per second: 943,349
 Total realtime factor: 4,716.74 x
 Total time per step: 1060.05 ns
 ```
 
 main:
 ```

 Total JIT time: 0.48 s
 Total simulation time: 24.11 s
 Total steps per second: 339,711
 Total realtime factor: 1,698.56 x
 Total time per step: 2943.68 ns
 ```
 
 **solver=newton, iterations=10, ls_iterations=20**
 ```
 mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton --iterations=10 --ls_iterations=20
 ```
 
 this pr:
 ```

 Total JIT time: 3.29 s
 Total simulation time: 23.92 s
 Total steps per second: 342,519
 Total realtime factor: 1,712.59 x
 Total time per step: 2919.55 ns
 ```
 
 main:
 ```

 Total JIT time: 0.47 s
 Total simulation time: 46.95 s
 Total steps per second: 174,468
 Total realtime factor: 872.34 x
 Total time per step: 5731.70 ns
 ```
 
 **for both solvers and multiple solver iterations, the early returns improve performance compared to the existing implementation**
 
 **in summary, the benchmarks demonstrate that early returns introduce a small amount of overhead for one solver iteration and can improve performance for multiple solver iterations.**

### CODE REVIEW

**Comment by erikfrey** [comment]
Let's discuss this in person along with your parallel linesearch PR - I think they're actually related and I want to talk over how we think about them and whether we would automatically switch between iterative and parallel, and how.

File: mujoco_warp/_src/io.py [hunk 1]
```diff
@@ -346,7 +346,7 @@ def _constraint(nv: int, nworld: int, njmax: int) -> types.Constraint:
   efc.beta = wp.empty(shape=(nworld,), dtype=wp.float32)
   efc.beta_num = wp.empty(shape=(nworld,), dtype=wp.float32)
   efc.beta_den = wp.empty(shape=(nworld,), dtype=wp.float32)
-  efc.done = wp.empty(shape=(nworld,), dtype=wp.int32)
+  efc.done = wp.empty(shape=(nworld,), dtype=bool)
```
erikfrey, line 349: nice, i approve

mujoco_warp/_src/solver.py
```diff
@@ -104,6 +114,16 @@ def _efc_kernel(d: types.Data):
     else:
       d.efc.force[efcid] = 0.0
 
+  @kernel
+  def _zero_qfrc_constraint(d: types.Data):
+    worldid, dofid = wp.tid()
+
+    if wp.static(m.opt.iterations) > 1:
+      if d.efc.done[worldid]:
```
erikfrey, line 122: Wow I'm kind of fascinated that even such a simple kernel benefits from early returning.

Not urgent but maybe we can bring this up with the Warp guys at some point, e.g. is there a good way to early return at the function level, otherwise we'll end up sprinkling this into so many kernels!

mujoco_warp/_src/solver.py
```diff
@@ -162,13 +208,23 @@ def _grad(d: types.Data):
     @kernel
     def _zero_h_lower(m: types.Model, d: types.Data):
       worldid, elementid = wp.tid()
+
+      if ITERATIONS > 1:
```
erikfrey, line 212: why is this one pulling `ITERATIONS` from currying instead of reading it off `m` like the other code?  If there's a good reason, let's add a comment because it's super unobvious to me.

If there's no particular reason, can we delete `ITERATIONS` and make this code read `m.opt.iterations` like the other kernels?

erikfrey (approved):
OK this looks great, just one nit and then let's get it merged.

mujoco_warp/_src/solver.py
thowell: there are cases where a `Model` variable is an argument to a kernel and then `wp.static` doesn't work with `m.opt.iterations`. here is an example error message:
```
Error evaluating static expression: name 'm' is not defined. Make sure all variables used in the static expression are constant
```
it is possible to not to use `wp.static` but this seems have a small but noticeable effect on performance.

added todos like this existing todo: https://github.com/erikfrey/mjx_warp/blob/e31f342b93574be713cac4e6c3e3e36087a4b08c/mujoco_warp/_src/solver.py#L288 to the kernels that utilize `ITERATIONS`

thowell: most recent commit includes the todos

mujoco_warp/_src/solver.py
thowell: not sure if there are performance benefits to early returning in all the cases, this is something to potentially investigate in the future.

PR #80, opened 2025-03-10
option to visualize mujoco c simulation

thowell:
potentially useful for debugging


mujoco/mjx/viewer.py
```diff
@@ -33,7 +33,9 @@
 _CLEAR_KERNEL_CACHE = flags.DEFINE_bool(
   "clear_kernel_cache", False, "Clear kernel cache (to calculate full JIT time)"
 )
-
+_IS_MJC = flags.DEFINE_bool(
```
erikfrey, line 36: let's do "--engine" with `flags.DEFINE_enum` with two modes to start: mjwarp, mjc.  Defaults to mjwarp.

We may add others in the future.

mujoco/mjx/viewer.py
```diff
@@ -57,42 +59,51 @@ def _main(argv: Sequence[str]) -> None:
...
-  print("Compiling the model physics step...")
-  start = time.time()
-  mjx.step(m, d)
-  # double warmup to work around issues with compilation during graph capture:
-  mjx.step(m, d)
-  # capture the whole smooth.kinematic() function as a CUDA graph
-  with wp.ScopedCapture() as capture:
+
+  if _IS_MJC.value:
+    m = mjm
```
erikfrey, line 64: nit: since we have to branch on the step function anyway, let's just use `mjm` and `mjd` for when we call `mujoco.mj_step`

so further down I would expect:

`mjx.step(m, d)`

`mujoco.mj_step(mjm, mjd)`

erikfrey (approved):
Looks great, just two nits.

mujoco/mjx/viewer.py
thowell: done

mujoco/mjx/viewer.py
thowell: done

PR #84, opened 2025-03-11
modify n^2 broadphase to count collisions by world

thowell:


thowell:
the modification to `nxn_broadphase` makes `broadphase_result_count` compatible with `group_contacts_by_type` when `nworld > 1`. 

`group_contacts_by_type` indexes `broadphase_result_count` by world here:
https://github.com/erikfrey/mjx_warp/blob/353c1f9969585863901132c063c4f52be703abe0/mujoco/mjx/_src/collision_driver.py#L353)

erikfrey (approved):
Got it, thank you!

PR #85, opened 2025-03-11
restructure solver memory

thowell:
```
mjx-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --solver=newton
```

this pr:
```

 Total JIT time: 0.18 s
 Total simulation time: 7.96 s
 Total steps per second: 1,029,715
 Total realtime factor: 5,148.57 x
 Total time per step: 971.14 ns
```

main:
```

 Total JIT time: 0.18 s
 Total simulation time: 8.11 s
 Total steps per second: 1,009,871
 Total realtime factor: 5,049.36 x
 Total time per step: 990.23 ns
```

### DIFF
--- a/mujoco/mjx/_src/io.py
+++ b/mujoco/mjx/_src/io.py
@@ -304,6 +304,59 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   return m
 
 
+def _constraint(nv: int, nworld: int, njmax: int) -> types.Constraint:
+  efc = types.Constraint()
+
+  efc.J = wp.zeros((njmax, nv), dtype=wp.float32)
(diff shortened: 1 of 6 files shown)


mujoco/mjx/_src/io.py
```diff
@@ -304,6 +304,59 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   return m
 
 
+def _constraint(nv: int, nworld: int, njmax: int) -> types.Constraint:
```
erikfrey, line 307: that's cool, maybe in the future we can construct contact this way too.

no action for now.

erikfrey (approved):
ship it!

PR #87, opened 2025-03-11
Lower ctrl to reasonable in-range domain that produces stable physics.

erikfrey:
Out of range ctrl creates large qacc which leads to test flakiness.


erikfrey:
Merging to get us back to green :heavy_check_mark:

PR #89, opened 2025-03-12
Jax unroll script and changes needed.

erikfrey:


adenzler-nvidia (approved):
thanks - no objections. Looking forward to trained robots!

erikfrey:
OK!  I think/hope the `kernel_copy` thing is temporary, so added a TODO to remove it so we don't forget.

PR #90, opened 2025-03-12
add scene to benchmark collision

thowell:
```
mjwarp-testspeed --function=collision --mjcf=collision.xml --batch_size=8192
```

```
nbody: 7
ngeom: 7
ncon: 35 
nefc: 108
```


mujoco_warp/test_data/collision/scene.xml
```diff
@@ -0,0 +1,25 @@
+<mujoco>
```
erikfrey, line 1: These bodies are all at origin yeah?  So this scene will explode as you step through it?

Does it impact what you're trying to test if we give these some offset positions so the scene doesn't explode?  These XMLs sometimes have a life beyond what we think they'll be used for and having generally stable physics for them is probably a good general goal.

erikfrey (approved):
Looks great, just one nit

mujoco_warp/test_data/collision/scene.xml
thowell: yes, all of the bodies are initially located at the origin and the scene will explode at the beginning of the simulation.

what are your thoughts about setting the timestep to zero in the xml? otherwise i can add a keyframe to set the initial system configuration.

erikfrey: a keyframe is OK, I would probably have just used <body pos="..." to keep it simple

thowell: - individual xml files are combined into a single file: `collision.xml` and replicate is no longer utilized
- body positions are initialized using `pos="..."`
- a keyframe is added that sets all bodies to the origin in order to collide all geoms

PR #94, opened 2025-03-13
Add proper plane support for the sweep and prune broad phase

nvtw:
Add proper plane support for the sweep and prune broad phase and switch to using bounding spheres similar as the n^2 broad phase does it. Going back to boxes is possible but spheres simplify the handling of plane interactions.
tile_sort support is in the MR too but deactivated since the feature is not yet in warp.

### DIFF
--- a/mujoco_warp/_src/broad_phase_test.py
+++ b/mujoco_warp/_src/broad_phase_test.py
@@ -25,26 +25,63 @@ import mujoco_warp as mjwarp
 
 from . import collision_driver
 from . import test_util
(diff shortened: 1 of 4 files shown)


nvtw:
Just noticed that the change broke the unit tests, will fix them now

erikfrey:
defer to @adenzler-nvidia but generally LGTM

adenzler-nvidia (approved):
LGTM - we'll see if we need AABBs again in the future but it's not a big thing to add them/have the broadphase be configurable.

PR #96, opened 2025-03-13
add function _add_geom_pair for broadphase routines

thowell:


adenzler-nvidia (approved):
makes a lot of sense, thank you!

PR #100, opened 2025-03-14
fix crb (dense) and add test

thowell:
fix crb (dense) and add test

```
mjwarp-testspeed --function=crb --mjcf=humanoid/humanoid.xml --batch_size=8192 --is_sparse=False
```

this pr:
```

 Total JIT time: 0.51 s
 Total simulation time: 0.48 s
 Total steps per second: 16,949,446
 Total realtime factor: 84,747.23 x
 Total time per step: 59.00 ns
```

main:
```

 Total JIT time: 0.01 s
 Total simulation time: 0.42 s
 Total steps per second: 19,680,173
 Total realtime factor: 98,400.86 x
 Total time per step: 50.81 ns
```

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -260,32 +260,32 @@ def crb(m: Model, d: Data):
     wp.atomic_add(d.crb, worldid, pid, d.crb[worldid, bodyid])
 
   @kernel
-  def qM_sparse(m: Model, d: Data, qM: wp.array(ndim=3, dtype=wp.float32)):
+  def qM_sparse(m: Model, d: Data):
     worldid, dofid = wp.tid()
     madr_ij = m.dof_Madr[dofid]
     bodyid = m.dof_bodyid[dofid]
 
     # init M(i,i) with armature inertia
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -260,38 +260,42 @@ def crb_accumulate(m: Model, d: Data, leveladr: int):
...
       madr_ij += 1
       dofid = m.dof_parentid[dofid]
 
   @kernel
-  def qM_dense(m: Model, d: Data):
+  def qM_dense(m: Model, d: Data, qM: wp.array(ndim=3, dtype=wp.float32)):
     worldid, dofid = wp.tid()
     bodyid = m.dof_bodyid[dofid]
 
-    # init M(i,i) with armature inertia
```
erikfrey, line 285: nit: let's keep this comment but update it

erikfrey (approved):
Thank you!

mujoco_warp/_src/smooth.py
thowell: restored the comment and reorganized `qM_dense` a bit

PR #103, opened 2025-03-14
modifications and updates for broadphase and tests

thowell:
this pr is focused on cleanup for broadphase and broadphase tests


erikfrey:
@thowell Can you add some description to your PR?  Is this just cleanup or are there functional changes?

mujoco_warp/_src/types.py
```diff
@@ -479,15 +479,15 @@ class Data:
...
-  boxes_sorted: wp.array(dtype=wp.vec3, ndim=3)
-  box_projections_lower: wp.array(dtype=wp.float32, ndim=2)
-  box_projections_upper: wp.array(dtype=wp.float32, ndim=2)
-  box_sorting_indexer: wp.array(dtype=wp.int32, ndim=2)
-  ranges: wp.array(dtype=wp.int32, ndim=2)
-  cumulative_sum: wp.array(dtype=wp.int32, ndim=1)
-  segment_indices: wp.array(dtype=wp.int32, ndim=1)
-  dyn_geom_aabb: wp.array(dtype=wp.vec3, ndim=3)
+  # sweep-and-prune broadphase
+  sap_geom_sort: wp.array(dtype=wp.vec3, ndim=3)
```
adenzler-nvidia, line 483: I would prefer that one to be sap_boxes_sort, but I'll leave the final word up to @nvtw.

adenzler-nvidia (approved):
Makes sense with the prefix. Thanks.

mujoco_warp/_src/types.py
thowell: how about `sap_box_sort`?

adenzler-nvidia: I think this is irrelevat now, the member does not exist anymore because Tobi switched the SAP broadphase to be sphere-based instead of AABB-based. So all good, feel free to merge.

PR #107, opened 2025-03-15
Remove event_scope from _linesearch.

erikfrey:
Prevents crashing when running mjwarp-testspeed with --event_trace=True.

I think this is because of nested traces from solve -> linesarch -> mul_m which the tracer does not handle correctly.


thowell:
not experiencing a crash with @event_scope using:
```
mjwarp-testspeed --function=step --mjcf=humanoid/humanoid.xml --batch_size=8192 --event_trace=True
```
(cuda toolkit 12.8, driver 12.2)

erikfrey:
Try raising iterations to 2?

erikfrey:
(in the testspeed params, not the xml)

thowell (approved):

PR #108, opened 2025-03-15
update types

thowell:
- update types based on currently supported/unsupported features
- add `disableflags` functionality
- add todos for tests


erikfrey (approved):
Très cool.

PR #110, opened 2025-03-15
Add support for actuator gain/bias used in position control.

erikfrey:


erikfrey:
I'll add tests next week.  Just racing to the finish line on features needed for the colab.

thowell (approved):
add todo comments in forward.py and forward_test.py?

PR #112, opened 2025-03-16
updates for primitive collisions

thowell:
- update collision_driver_test.py
- rename: collision_functions.py -> collision_primitive.py
- rename narrowphase function for primitives to `primitive_narrowphase`
- remove imports from `collision` function


erikfrey:
File renaming makes sense (to match MuJoCo) but defer to @adenzler-nvidia in case there was a good reason to name it `collision_functions.py`

adenzler-nvidia:
no good reason, that was the first thing I came up with. Your call on how much to match to MuJoCo - in PhysX/Nvidia land primitives are usually all geometries except meshes, sdf and heightfields. So I'm not sure if that naming is totally correct if we would add mesh-mesh or other complex collision functions in there, but we can leave that for later. Even more now that it's unclear which of the collisions will be handled by the GJK code down the line.

thowell:
mujoco is organized by:
- box
- convex
- gjk
- primitive
- sdf

https://github.com/google-deepmind/mujoco/tree/main/src/engine

erikfrey:
OK, I think that's compatible with what @adenzler-nvidia is saying but let's wait for him to reply.

adenzler-nvidia:
sounds about right with the split. I think we can adapt this as we go - depends a bit on the plans of potentially moving the collisions to a separate package alltogether. In that light, I would just go for something intuitive that already makes sense, and following mujoco is probably the right thing.

If we decide to go ahead with a separate collision module, I envision this to be a toolbox kind of plug-and-play thing that has lots of options and configuration knobs where basically every algorithm that is somehow imaginable is GPU accelerated, and every engine can decide which parts to pick and implement. Just thinking while we go along here.

erikfrey:
That would be a very cool place to end up!  I could definitely see collisions being a universally useful library.

OK sounds like we are good to go here and can merge once you fix conflicts.

erikfrey (approved):

PR #113, opened 2025-03-17
Colab to train Apptronik Apollo.

erikfrey:


adenzler-nvidia (approved):
LGTM - I did not go through the colab itself in detail, I'm not an expert there. But the changes to Mujoco-warp look good.

PR #116, opened 2025-03-18
Fix typo.

erikfrey:


mujoco_warp/_src/forward.py
```diff
@@ -674,7 +674,7 @@ def forward(m: Model, d: Data):
   # TODO(team): sensor.sensor_acc
 
   if d.njmax == 0:
-    kernel_copy(d.qcc, d.qacc_smooth)
+    kernel_copy(d.qacc, d.qacc_smooth)
```
Kenny-Vilella, line 677: More generally, this condition will very rarely occurs.
The best will be to not run the solver if there is no active constraints.
But the number of active constraint is stored in a warp array that cannot be accessed in CPU code.
Not sure what is the best way to do this.

thowell (approved):

mujoco_warp/_src/forward.py
erikfrey: @thowell has built up some “don’t run the solver” experience with his recently landed early return code.

@thowell ia if possible we are missing some conditions?  Maybe you can load up the leap hand scene and try it out?  Or some other way of reproing that we have problems when d.nefc == 0?

